PR #6, opened 2022-09-07
restructure data and update contracts list

ujjwal-cyph:
Fixes https://app.clickup.com/t/37308523/CHI-1940
Fixes https://app.clickup.com/t/37308523/CHI-1859


src/controller_main.c
```diff
@@ -573,14 +573,18 @@ void desktop_listener_task(lv_task_t* data)
...
                     flow_level.show_desktop_start_screen = true;
-                    // 29 is the size of the purpose, coin index, account index, chain index, address index, token name and network chain id
-                    memcpy(&receive_transaction_data, data_array , offset+ 29);
-                    offset+=29;
 
-                    coin_index = BYTE_ARRAY_TO_UINT32(receive_transaction_data.coin_index);
+                    uint16_t offset = byte_array_to_recv_txn_data(&receive_transaction_data,data_array,msg_size);
+
+                    if(offset == -1){
+                        comm_reject_request(RECV_TXN_USER_VERIFIED_COINS,0);
```
ujjwal-cyph, line 581: This indicates that the user has rejected the request.

see case `RECV_TXN_FIND_XPUB` inside `receive_transaction_controller_b` file.

ujjwal-cyph: I suggest, instead we define a new rejection case or use the generic rejection of "invalid command type" using a call to `comm_reject_invalid_cmd`.

src/controller_main.c
```diff
@@ -573,14 +573,18 @@ void desktop_listener_task(lv_task_t* data)
...
                     CY_Reset_Not_Allow(false);
-                    uint16_t offset = WALLET_ID_SIZE;
-                    uint32_t coin_index;
                     flow_level.show_desktop_start_screen = true;
-                    // 29 is the size of the purpose, coin index, account index, chain index, address index, token name and network chain id
-                    memcpy(&receive_transaction_data, data_array , offset+ 29);
-                    offset+=29;
 
-                    coin_index = BYTE_ARRAY_TO_UINT32(receive_transaction_data.coin_index);
+                    uint16_t offset = byte_array_to_recv_txn_data(&receive_transaction_data,data_array,msg_size);
```
ujjwal-cyph, line 578: return types do not match. checks will fail

src/controller_main.c
```diff
@@ -573,14 +573,18 @@ void desktop_listener_task(lv_task_t* data)
...
+
+                    if(offset == -1){
+                        comm_reject_invalid_cmd();
+                        reset_flow_level();
```
ujjwal-cyph, line 582: this is not needed. Infact, it is interfering with UI focus.

Test command used: `Sent: 59,1CBEBD0EC4590E39359A8073BE1B3F9DA5117144B0663986016AE442750815610000`

common/coin_support/coin_utils.h
```diff
@@ -108,17 +108,50 @@ typedef struct
...
+typedef struct Receive_Transaction_Data {
+  uint8_t wallet_id[WALLET_ID_SIZE];
+  uint8_t purpose[4];
+  uint8_t coin_index[4];
+  uint8_t account_index[4];
+  uint8_t chain_index[4];
+  uint8_t address_index[4];
+  char *token_name;
+  union {
+    uint8_t network_chain_id;
```
ujjwal-cyph, line 140: We will add new EVM chains. Would it be good to expand this now? Chain IDs can be multi-byte value https://chainlist.org/.

If we do this, we can remove the union.

TejasvCypherock: We can do this, however, this would delay the build as it requires changes from both sides.

ujjwal-cyph: We will take this up in later discussions

PR #17, opened 2022-09-19
Fix the fee calculation for non-integer Gwei gasPrice

ujjwal-cyph:
Currently the gas price is assumed to be of integer value in Gwei. If the value is non-integer, then the current method of calculation leads to truncation of values when converted to Gwei

Fixes https://app.clickup.com/t/37308523/CHI-2003


amanCypherock:
Add the logic for conversion of fee string to value in a function so it would be easier to run functional tests

PR #19, opened 2022-09-20
Bugfix: Reset passphrase before using

ghost:
https://app.clickup.com/t/37308523/CHI-2009

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -28,11 +28,10 @@
                 "cwd": "${workspaceRoot}/build"
             },
             "dependsOrder": "sequence",
-            "dependsOn": [
-                "config-main-simulator"
-            ],
+            "dependsOn": ["config-main-simulator"],
             "group": "build"
         },
+
         {
             "label": "config-initial-simulator",
             "type": "shell",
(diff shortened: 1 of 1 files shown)


amanCypherock:
Should add a check if a passphrase is used in the wallet or not when converting from mnemonic to seed. Based on that we should either pass a passphrase or an empty string.

amanCypherock:
Also, clear the passphrase in all flows on flow resets, if not done already

src/level_four/core/controller/receive_transaction_controller.c
```diff
@@ -75,6 +75,7 @@ void receive_transaction_controller()
     switch (flow_level.level_three) {
 
     case RECV_TXN_FIND_XPUB: {
+        memzero(wallet_credential_data.passphrase, sizeof(wallet_credential_data.passphrase));
```
ujjwal-cyph, line 78: lets reset this in `reset_flow_level` as well.

ujjwal-cyph: Also, add this just after passphrase is used. i.e., after the call to `mnemonic_to_seed`

ujjwal-cyph (approved):

src/controller_main.c
```diff
@@ -229,6 +229,8 @@ void reset_flow_level()
 
     //clear memory
     memzero(wallet.password_double_hash, sizeof(wallet.password_double_hash));
+    memzero(wallet_credential_data.passphrase, sizeof(wallet_credential_data.passphrase));
+
```
ujjwal-cyph, line 233: no space. `cy_free` belongs to same action group i.e., cleaning memory allocations

ujjwal-cyph, line 233:
```suggestion
```

PR #22, opened 2022-09-21
Increase the defined max word length to 16 to match wallet names

ujjwal-cyph:
The ui_list component is defined with rows size of 15 in one of its parameter. Using the component to list wallet names results in trimming of last letter of the wallet name.

Fixes https://app.clickup.com/t/37308523/CHI-2010


common/coin_support/wallet.h
```diff
@@ -42,13 +42,13 @@
 #define MAX_NUMBER_OF_MNEMONIC_WORDS 24
 
 /// Max length of mnemonic word
-#define MAX_MNEMONIC_WORD_LENGTH 16
+#define MAX_MNEMONIC_WORD_LENGTH 9
```
ujjwal-cyph, line 45: need same size as the `ui_list` function parameter i.e. `MAX_UI_LIST_CHAR_LEN`.

amanCypherock:
```suggestion
#define MAX_MNEMONIC_WORD_LENGTH MAX_UI_LIST_CHAR_LEN
```

vbcypherock:
Task linked: [CHI-2010 Update ui_list component to allow 16 characters long entries](https://app.clickup.com/t/37308523/CHI-2010)

common/interfaces/user_interface/ui_list.h
```diff
@@ -12,7 +12,10 @@
 
 #include "ui_common.h"
 
-//TODO : Add constant
+//TODO: Update count for higher coin list
+#define MAX_UI_LIST_WORDS       24
+#define MAX_UI_LIST_CHAR_LEN    16
```
ujjwal-cyph, line 17: The limit of 15 on length of each entry to the list is not useful. The add coin flow lists the coin names which could be larger.

amanCypherock: To solve this issue we can use the following solution
- Make option_list a array of pointer instead of a double array
- use wrapper for mnemonic implementation and different API other uses of ui_list which take array of pointer as input.

amanCypherock: This change will be picked up in https://github.com/Cypherock/x1_wallet_firmware/pull/221

amanCypherock (approved):

PR #25, opened 2022-09-24
Increase the defined max word length to 16 to match wallet names

ujjwal-cyph:
The ui_list component is defined with rows size of 15 in one of its parameter. Using the component to list wallet names results in trimming of last letter of the wallet name.

Fixes https://app.clickup.com/t/37308523/CHI-2010

### DIFF
--- a/common/coin_support/wallet.h
+++ b/common/coin_support/wallet.h
@@ -42,7 +42,7 @@
 #define MAX_NUMBER_OF_MNEMONIC_WORDS 24
 
 /// Max length of mnemonic word
-#define MAX_MNEMONIC_WORD_LENGTH 15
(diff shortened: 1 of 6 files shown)


ujjwal-cyph:
Test conversation rules

common/coin_support/wallet.h
```diff
@@ -42,7 +42,7 @@
 #define MAX_NUMBER_OF_MNEMONIC_WORDS 24
 
 /// Max length of mnemonic word
-#define MAX_MNEMONIC_WORD_LENGTH 15
+#define MAX_MNEMONIC_WORD_LENGTH 16
```
ujjwal-cyph, line 45: test conversation rules

PR #41, opened 2022-10-08
Show fast tap message if card absent before start of pairing

ujjwal-cyph:
Fixes https://app.clickup.com/t/37308523/CHI-2030

### DIFF
--- a/common/interfaces/card_interface/nfc.c
+++ b/common/interfaces/card_interface/nfc.c
@@ -64,7 +64,6 @@
 #include "application_startup.h"
 #include "app_error.h"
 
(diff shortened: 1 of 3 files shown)


ujjwal-cyph:
The message appears even in normal flows wherein card is actually present.

vbcypherock:
Task linked: [CHI-2030 Check card presence before moving for pairing](https://app.clickup.com/t/37308523/CHI-2030)

ujjwal-cyph:
> The message appears even in normal flows wherein card is actually present.

This is fixed now by using `tg-init` instead of `card-presence-detection`

amanCypherock (approved):

PR #53, opened 2022-10-28
add new command to get latest blockhash

TejasvCypherock:


vbcypherock:
Task linked: [CHI-2058 BugFix: Receive blockhash in Solana send flow before signing](https://app.clickup.com/t/37308523/CHI-2058)

ujjwal-cyph (changes requested):

src/level_four/core/controller/send_transaction_controller_solana.c
```diff
@@ -175,7 +175,31 @@ void send_transaction_controller_solana() {
     } break;
 
     case SEND_TXN_TAP_CARD_SEND_CMD_SOLANA: {
-      flow_level.level_three = SEND_TXN_READ_DEVICE_SHARE_SOLANA;
+      flow_level.level_three = SEND_TXN_UPDATE_BLOCKHASH_SOLANA;
+      uint8_t arr[1]         = {1};
+      transmit_data_to_app(SEND_TXN_PRE_SIGNING_DATA, arr, sizeof(arr));
```
amanCypherock, line 180: Would prefer to move this command response to a task before the pin entry.
You can check the flow_status as a trigger for sending blockhash

src/level_four/core/controller/send_transaction_controller_solana.c
```diff
@@ -175,7 +175,31 @@ void send_transaction_controller_solana() {
...
+      if (get_usb_msg_by_cmd_type(SEND_TXN_PRE_SIGNING_DATA, &blockhash_received, &blockhash_msg_size)) {
+        if (blockhash_msg_size != SOLANA_BLOCKHASH_LENGTH) {
+          comm_reject_invalid_cmd();
+          reset_flow_level();
+        }
+        memcpy(solana_latest_blockhash, blockhash_received, SOLANA_BLOCKHASH_LENGTH);
+        clear_message_received_data();
+
+        int status = solana_update_blockhash_in_byte_array(solana_unsigned_txn_byte_array, solana_latest_blockhash);
+        if (status != SEC_OK) {
```
amanCypherock, line 197: SEC should be replaced with SOL overall for solana error codes.
Either add change in this MR or raise another

src/level_four/core/controller/send_transaction_controller_solana.c
```diff
@@ -175,7 +175,31 @@ void send_transaction_controller_solana() {
...
+          LOG_ERROR("Solana error code: %d\n BlockHash not updated", status);
```
amanCypherock, line 198: Should reject this as a failure for the following reasons
1. This flow depends on this blockhash, if it cannot be updated probability of failure is quite high
2. Desktop would not get a feedback that command had actually failed

src/level_four/core/controller/send_transaction_controller_solana.c
amanCypherock: Also `SEND_TXN_PRE_SIGNING_DATA` is not the right response code. Device should send response to the previous command id `SEND_TXN_REQ_UNSIGNED_TXN`

amanCypherock (changes requested):
Changes required regarding command type, response sequence and other issues

src/level_four/core/controller/send_transaction_controller_solana.c
TejasvCypherock: Actually, the desktop will show the error for invalid signature as it is expecting the signature for a different blockhash

src/level_four/core/controller/send_transaction_controller_solana.c
TejasvCypherock: Updated the command Id but couldn't move the send command as the desktop still expects pin / passphrase/ card tap rejection

src/level_four/core/controller/send_transaction_controller_solana.c
```diff
@@ -175,7 +175,33 @@ void send_transaction_controller_solana() {
...
+        if (blockhash_msg_size != SOLANA_BLOCKHASH_LENGTH) {
+          comm_reject_invalid_cmd();
+          reset_flow_level();
+        }
+        memcpy(solana_latest_blockhash, blockhash_received, SOLANA_BLOCKHASH_LENGTH);
+        clear_message_received_data();
+
+        int status = solana_update_blockhash_in_byte_array(solana_unsigned_txn_byte_array, solana_latest_blockhash);
+        if (status != SOL_OK) {
+          LOG_ERROR("Solana error code: %d\n BlockHash not updated", status);
```
ujjwal-cyph, line 198: shorten this error message. keeping the status value as it is.

ujjwal-cyph (approved):

amanCypherock (approved):

PR #90, opened 2022-12-13
fix: Correct the USB state on timeout

ujjwal-cyph:
Fixes https://app.clickup.com/t/37308523/CHI-2120

### DIFF
--- a/common/startup/application_startup.c
+++ b/common/startup/application_startup.c
@@ -163,10 +163,10 @@ static void repeated_timer_handler(void)
         inactivity_counter = 0;
         if (counter.level > LEVEL_ONE) {
             mark_error_screen(ui_text_process_reset_due_to_inactivity);
-            reset_flow_level();
-            lv_obj_clean(lv_scr_act());
             if (CY_External_Triggered())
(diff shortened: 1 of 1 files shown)


vbcypherock:
Task linked: [CHI-2120 Correct state handling on timeout](https://app.clickup.com/t/37308523/CHI-2120)

common/startup/application_startup.c
```diff
@@ -163,10 +163,10 @@ static void repeated_timer_handler(void)
         inactivity_counter = 0;
         if (counter.level > LEVEL_ONE) {
             mark_error_screen(ui_text_process_reset_due_to_inactivity);
-            reset_flow_level();
-            lv_obj_clean(lv_scr_act());
             if (CY_External_Triggered())
                 comm_reject_request(STATUS_PACKET, STATUS_CMD_ABORT);
+            reset_flow_level();
```
amanCypherock, line 168: This fix might be needed in other areas where reset flow level is called before external triggered check.
Should identify cases and add those too.

ujjwal-cyph: verified. no other such instance.

amanCypherock (approved):

PR #97, opened 2022-12-21
fix: Add missing image in markdown documentation

ujjwal-cyph:


amanCypherock (changes requested):
Few changes required
1. Update screen names to correct names (e.g. "sync with cycard" should be replaced with Restore from Cards)
2. Settings flow level are different from actual, could replace ... as there are many screens in settings
3. Dotted lines are inconsistent and confusing
4. View seed dummy is incorrect

amanCypherock (approved):

PR #98, opened 2022-12-21
feat: Enable use of multi-account per wallet

ujjwal-cyph:
Fixes https://app.clickup.com/t/37308523/CHI-2050


vbcypherock:
Task linked: [CHI-2050 Implement the changes on device and test with CLI](https://app.clickup.com/t/37308523/CHI-2050)

src/level_four/core/controller/add_coin_controller.c
```diff
@@ -147,41 +147,19 @@ void add_coin_controller()
...
-                generate_xpub(path, 3, SECP256K1_NAME, seed, (char *) cmd_add_coin.xpubs[j]);
-                memzero(path, sizeof(path));
-            }
-            if (add_coin_data.coin_indexes[i] == 0x80000000 || add_coin_data.coin_indexes[i] == 0x80000001) {
-                uint32_t path[] = { NATIVE_SEGWIT, add_coin_data.coin_indexes[i], 0x80000000 };
-                generate_xpub(path, 3, SECP256K1_NAME, seed, (char *) cmd_add_coin.xpubs[++j]);
-                memzero(path, sizeof(path));
-            }
-        }
+        generate_xpub(add_coin_data.derivation_path, add_coin_data.derivation_depth, curve, seed, (char *) cmd_add_coin.xpub);
```
amanCypherock, line 158: Need coin specific checks on path before generating xpubs

common/coin_support/coin_utils.h
```diff
@@ -132,6 +148,7 @@ typedef struct
     uint64_t network_chain_id;
 
     uint8_t is_harmony_address;
+    uint16_t address_tag;
```
amanCypherock, line 151: Is the `address_tag` a generic variable that could be used in other chain paths too?
If `address_tag` is only useful for solana, then we should rename it as `sol_address_tag`.

We could also have a union variable of size 64 bytes or more, which could be used to populate any chain specific variables like `address_tag` and `is_harmony_address` and not needing to update the txn_metadata.

amanCypherock (changes requested):
Need to update text for receive and send transaction to include address tag i.e. segwit, native, legacy and account no.

common/coin_support/coin_utils.h
ujjwal-cyph: this is going to be generic. not specific to solana. currently its defined only for solana. but can be used for others as well.

src/level_four/core/controller/add_coin_controller.c
ujjwal-cyph: added

ujjwal-cyph:
> xt for receive and send transaction to include address tag i.e. segwit, native, legacy and accou

added

amanCypherock (changes requested):
Update base to develop

common/coin_support/coin_utils.c
```diff
@@ -456,4 +489,111 @@ void bech32_addr_encode(char *output, char *hrp, uint8_t *address_bytes, uint8_t
...
+            break;
+
+        default:
+            break;
+    }
+
+    return status;
+}
+
+bool verify_receive_derivation_path(const uint32_t *path, uint8_t depth) {
```
amanCypherock, line 528: Are we planning to use this function in future?

ujjwal-cyph: thanks for reminding. i have to add this in receive address flow.

ujjwal-cyph:
merged `develop` and updated base to `develop`

src/controller_main.c
```diff
@@ -541,39 +536,68 @@ void desktop_listener_task(lv_task_t* data)
...
                         clear_message_received_data();
                         comm_reject_invalid_cmd();
                         return;
                     }
+                    uint32_t path[5]    = {
+                        BYTE_ARRAY_TO_UINT32(receive_transaction_data.purpose),
+                        BYTE_ARRAY_TO_UINT32(receive_transaction_data.coin_index),
+                        BYTE_ARRAY_TO_UINT32(receive_transaction_data.account_index),
+                        BYTE_ARRAY_TO_UINT32(receive_transaction_data.chain_index),
+                        BYTE_ARRAY_TO_UINT32(receive_transaction_data.address_index)};
```
ujjwal-cyph, line 553: use send metadata instead of receive struct

common/coin_support/coin_utils.h
amanCypherock, line 151:
```suggestion
    uint16_t address_tag;    //Used to differentiate address paths of the same chains
```

common/coin_support/coin_utils.c
```diff
@@ -456,4 +489,111 @@ void bech32_addr_encode(char *output, char *hrp, uint8_t *address_bytes, uint8_t
...
+    uint16_t length = 0;
+    char *type = "";
+
+    switch (coin) {
+        case NEAR:
+            length = snprintf(account_name, out_len, "idx.%lu", (near_get_account_index(path) + 1));
+            break;
+
+        case SOLANA:            // m/44'/501'/i'/j'
+            type = account_type == 1 ? "sol-base" : (account_type == 2) ? "type1" : "type2";
```
ujjwal-cyph, line 577:
```suggestion
            type = account_type == 1 ? "paper" : (account_type == 2) ? "ledger" : "phantom";
```

common/coin_support/near.c
```diff
@@ -183,4 +183,21 @@ size_t near_get_account_ids_count(const uint8_t* data,const uint16_t data_len){
...
+
+  // m/44'/397'/0'/0'/i'
+  status = (purpose == NON_SEGWIT && coin == NEAR && account == 0x80000000 && change == 0x80000000);
+
+  return status;
+}
+
+uint32_t near_get_account_index(const uint32_t *path) {
+  // discard the sign bit denoting hardened/non-harndedned value
+  return ((path[4] & 0x7FFFFFFF) - 1);
```
ujjwal-cyph, line 202: earlier assumption was that accounts starts from index 1. now indices can start from 0 as well.

```suggestion
  return (path[4] & 0x7FFFFFFF);
```

amanCypherock (approved):

PR #100, opened 2022-12-22
fix: Allow coin list query in unauthenticated device

ujjwal-cyph:
Fixes https://app.clickup.com/t/37308523/CHI-2131


vbcypherock:
Task linked: [CHI-2131 Allow coin list even when device is unauthenticated](https://app.clickup.com/t/37308523/CHI-2131)

ujjwal-cyph:
Not required. This will be handled on desktop application.

PR #118, opened 2023-01-09
Use exact buffer size for ethereum addresses

ujjwal-cyph:
Ethereum default address of the form `0xe688b84b23f322a994a53dbf8e15fa82cdb71127` is of 43 byte including the `null` character.

Also, harmony's bech32 addresses also result into 43 bytes including `null` character.

So size 43 fits both the types of addresses.


amanCypherock:
Have you tested with the change?

amanCypherock (approved):

PR #149, opened 2023-02-06
PRF-4050: update text to inform user that PIN can be alphanumeric during wallet creation

kartikCypherock:
…ng wallet creation


rohanagarwal94:
Task linked: [PRF-4050 On device in setting pin or entering pin, the next character always lands on Numeric pin instead of a mix of alphabets and numbers](https://app.clickup.com/t/9002019994/PRF-4050)

PR #152, opened 2023-02-11
fix: add logs and increase memory for lvgl

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4521 BugFix: lvgl memory size increase](https://app.clickup.com/t/9002019994/CHI-4521)

common/startup/application_startup.c
```diff
@@ -358,6 +358,7 @@ void application_init() {
         ui_rotate();
     }
     logger_init();
+    lv_log_register_print_cb((lv_log_print_g_cb_t)lv_custom_log_handler);
```
ujjwal-cyph, line 361: This explicit typecasting should not be needed.

common/logger/logger.c
```diff
@@ -319,3 +318,9 @@ log_read_e_t get_log_read_status()
 {
     return sg_log_data.read_sm_e;
 }
+
+void lv_custom_log_handler(lv_log_level_t level, const char *file, int line, const char *dsc) {
```
ujjwal-cyph, line 322: the function prototype is wrong. Correction: 
```suggestion
void lv_custom_log_handler(lv_log_level_t level, const char *file, uint32_t line, const char *dsc) {
```

ujjwal-cyph (changes requested):

common/lv_conf.h
```diff
@@ -223,7 +223,7 @@ typedef void * lv_indev_drv_user_data_t;            /*Type of user data in the i
  *===============*/
 
 /*1: Enable the log module*/
-#define LV_USE_LOG      0
+#define LV_USE_LOG      1
```
ujjwal-cyph, line 226: this bumps the firmware binary size by `3256 bytes`. we might want to do this by different mechanism (directly modifying the `lv_mem_assert` macro in `lv_mem.h`.

ujjwal-cyph (approved):

PR #169, opened 2023-02-25
Clear structures before decoding unsigned transactions

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4554 Bugfix: Clear structures before decoding unsigned transactions](https://app.clickup.com/t/9002019994/CHI-4554)

amanCypherock (changes requested):
Should include this for APIs that populate data from byte array to other structures.

There are 3 such APIs in `coin_utils.c` and for `eth.c` the new API `eth_byte_array_to_msg`

TejasvCypherock:
> 

`eth_byte_array_to_msg` already has an explicit init funciton `eth_init_msg_data` which is being called before receiving the message in the flow.

amanCypherock (approved):

PR #172, opened 2023-02-28
zero-ise buffer before reading back io protection key and ext keys

kartikCypherock:
Fixes comparison of keys installed on device: zero-ise key buffer before reading back from flash memory


rohanagarwal94:
Task linked: [CHI-4340 Clear buffers before reading back the written values for compare](https://app.clickup.com/t/9002019994/CHI-4340)

amanCypherock (approved):

PR #179, opened 2023-03-06
Emulate key presses for dev build

TejasvCypherock:


common/libraries/util/dev_utils.h
```diff
@@ -0,0 +1,30 @@
...
+  struct ekp_queue_node *next;
+} ekp_queue_node;
+
+typedef struct ekp_queue {
+  ekp_queue_node *front;
+  ekp_queue_node *rear;
+  int count;
+} ekp_queue;
+
+ekp_queue_node *ekp_new_queue_node(const lv_event_t event, const uint32_t delay);
```
amanCypherock, line 21:
```suggestion
/* Utils to Emulate key presses */
ekp_queue_node *ekp_new_queue_node(const lv_event_t event, const uint32_t delay);
```

amanCypherock: Also add some comments on how to use this feature while testing.

amanCypherock (approved):

PR #181, opened 2023-03-13
Remove usage of decimal from metadata for Ethereum tokens

TejasvCypherock:


common/coin_support/eth.c
```diff
@@ -408,37 +419,34 @@ uint32_t eth_get_value(const eth_unsigned_txn *eth_unsigned_txn_ptr, char *value
...
+             (cy_read_be(eth_utxn_ptr->chain_id, eth_utxn_ptr->chain_id_size[0]) != metadata_ptr->network_chain_id) || // Check if the chain id from metadata matches with the chain id from the unsigned transaction
+             (metadata_ptr->is_token_transfer && eth_is_token_whitelisted && !is_zero(eth_utxn_ptr->value,eth_utxn_ptr->value_size[0])) || // Check if token transfer is triggered with whitelisted token and amount is non zero
+             (eth_utxn_ptr->payload_status == PAYLOAD_CONTRACT_INVALID)); // Check if the payload status is invalid
 }
 
-static PAYLOAD_STATUS eth_decode_txn_payload(const eth_unsigned_txn *eth_utxn_ptr, const txn_metadata *metadata_ptr) {
+static PAYLOAD_STATUS eth_decode_txn_payload(const eth_unsigned_txn *eth_utxn_ptr, txn_metadata *metadata_ptr) {
     PAYLOAD_STATUS result = PAYLOAD_ABSENT;
+    eth_is_token_whitelisted = false;
     if (eth_utxn_ptr->payload_size > 0) {
```
kartikCypherock, line 433: What is `payload_size` is les than 32 bits? i.e. `< 4` bytes

common/coin_support/eth.c
```diff
@@ -808,3 +823,8 @@ void eth_derivation_path_to_string(const txn_metadata *txn_metadata_ptr, char *o
 uint8_t eth_get_decimal(const txn_metadata *txn_metadata_ptr) {
   return txn_metadata_ptr->is_token_transfer ? txn_metadata_ptr->eth_val_decimal[0] : ETH_DECIMAL;
 }
+
+const char *eth_get_address_title(const eth_unsigned_txn *eth_unsigned_txn_ptr) {
+  return (eth_unsigned_txn_ptr->payload_status != PAYLOAD_ABSENT && !eth_is_token_whitelisted) ? ui_text_verify_contract
```
kartikCypherock, line 828: Please break down complex statement using if/else block, or use parenthesis around complete statement.

src/level_four/core/controller/send_transaction_controller.c
```diff
@@ -84,7 +84,7 @@ Send_Transaction_Data var_send_transaction_data = {
         .input_count = {0}, .input = NULL,
         .output_count = {0}, .output = NULL,
         .change_count = {0}, .change = NULL,
-        .transaction_fees = {0}, .eth_val_decimal = {0},
```
kartikCypherock, line 87: we can initialize `eth_val_decimal` with `0xff` as well

common/coin_support/eth.c
```diff
@@ -499,7 +507,7 @@ void eth_init_msg_data(MessageData *msg_data) {
 }
 
 int eth_byte_array_to_unsigned_txn(const uint8_t *eth_unsigned_txn_byte_array,
-                                   size_t byte_array_len, eth_unsigned_txn *unsigned_txn_ptr, const txn_metadata *metadata_ptr)
+                                   size_t byte_array_len, eth_unsigned_txn *unsigned_txn_ptr, txn_metadata *metadata_ptr)
```
amanCypherock, line 510: Why have we made a change in the function prototype when there's no change in the function?

TejasvCypherock: The function responsible for decoding payload now updates the decimal value on txn_metadata which is being called by this function.

amanCypherock (changes requested):
While testing an issue was found, 
The address verification screen is not synchronized correctly with cySync in the case of a wallet with a passphrase.
Steps to recreate
- Initiate send token txn for any non-whitelisted token with a wallet with passphrase
- Accept all options till address screen.
 On cySync address screen moves to passphrase input screen which is out of sync with device.

common/coin_support/eth.c
kartikCypherock: I see that you have added a check for payload_size in statement below, so resolving this

kartikCypherock (approved):

amanCypherock (approved):

PR #183, opened 2023-03-18
Proper casting for conversion macros

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4581 Bugfix: Use proper casting in conversion macros](https://app.clickup.com/t/9002019994/CHI-4581)

common/libraries/util/utils.h
```diff
@@ -28,13 +28,13 @@
 /// Convert bit array of size 4 to uint32
 #define BYTE_ARRAY_TO_UINT32(x) ((x)[0] << 24 | (x)[1] << 16 | (x)[2] << 8 | (x)[3])
 /// Read 16-bit value from big-endian serialized byte-array
-#define U16_READ_BE_ARRAY(x) ((x)[0] << 8 | (x)[1])
+#define U16_READ_BE_ARRAY(x) ((uint16_t)(x)[0] << 8 | (x)[1])
```
kartikCypherock, line 31: 1. This is confusing as it is not clear what expression we are trying to cast to uint16_t

```suggestion
#define U16_READ_BE_ARRAY(x) (((uint16_t)((x)[0]) << 8) | (x)[1])
```

2. The second operand (```(x)[1]```) is internally promoted to ```uint16_t``` by compiler. I will recommend you to go through type promotion in C, to understand the same if not done already.
Starting point: https://www.emertxe.com/blog/type-promotion-in-c/#:~:text=Type%20promotion%20in%20C%20is,conversion%20and%20Explicit%20type%20conversion.

kartikCypherock: Hi, I checked the C operator precedence table (https://en.cppreference.com/w/c/language/operator_precedence)
According to this, type case has higher priority than `or` operation, therefore we don't require this change.

kartikCypherock:
Based on offline tests performed using these macros, I am approving the PR

kartikCypherock (approved):

PR #190, opened 2023-03-27
Check chain Id for tokens and add 0x before the address in clear signing

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4594 Bugfix: Check chain Id for tokens and add 0x before the address in clear signing](https://app.clickup.com/t/9002019994/CHI-4594)

kartikCypherock (approved):

PR #191, opened 2023-03-27
Use get_coin_symbol to compare symbol of parent coin

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4595 Bugfix: Correct condition to check token transfer](https://app.clickup.com/t/9002019994/CHI-4595)

PR #192, opened 2023-03-27
Skip amount screen if value is 0 and payload is present

TejasvCypherock:


rohanagarwal94:
Task linked: [CHI-4596 Skip the amount screen if the amount is zero and the payload is present](https://app.clickup.com/t/9002019994/CHI-4596)

amanCypherock (approved):

PR #193, opened 2023-03-27
feat (ci): Add clang-tidy checks on pull requests

ujjwal-cyph:
Currently this will only create a result of the check. We will enforce the clang-tidy suggestions at a later point.

The clang-tidy in CI will only check for files which have been modified in the concerned pull request. It is possible to run the check locally by using the bash script that is used by CI.


amanCypherock:
Please add a description

PR #195, opened 2023-03-27
feat (git-hook): Add pre-commit hook

ujjwal-cyph:
The hook enforces following rules:
- Perform clang-format changes
- Prevent any non-ASCII filenames
- Whitespace error check

### DIFF
--- a/utilities/install-hooks.sh
+++ b/utilities/install-hooks.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+chmod +x utilities/hooks/*
+cp -r utilities/hooks .git/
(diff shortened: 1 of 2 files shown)


utilities/install-hooks.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+chmod +x utilities/hooks/*
```
kartikCypherock, line 3: suggestion: can we somehow run this shell script from main build.sh? in that case we will not have to do anything manually

ujjwal-cyph: it is doable but, `build.sh` would not be the best place.

PR #200, opened 2023-04-04
Fix clean build script

ujjwal-cyph:


utilities/build.sh
```diff
@@ -114,4 +124,7 @@ fi
 
 # exit if configuration failed with errors
 if [ ! $? -eq 0 ]; then exit 1; fi
+if [[ ${clean_flag} = "true" ]]; then
+  "${BUILD_TOOL}" clean;
```
kartikCypherock, line 143: why do we need `;` at the end of line?
it is not needed when invoking for compilation (`${BUILD_TOOL}" -j8 all`)

utilities/build.sh
```diff
@@ -11,6 +12,15 @@ ACTIVE_TYPE=Main
 BUILD_TYPE=Debug
 BUILD_PLATFORM=Device
 
+while getopts 'c' flag; do
```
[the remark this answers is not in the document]
ujjwal-cyph: The getopts is an internal flag parser. This will add flexibility for parameter ordering and optionality.

For example, the parameters of the script are optional `[main|initial] [dev|debug|release] [device|simulator]` but the following execution is wrong.
```
./build.sh release simulator
```

So using `getopts` will actually make the optional arguments.

utilities/build.sh
ujjwal-cyph: fixed

utilities/build.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/sh
 
 usage () {
-    echo -e "\tUSAGE: $0 [main|initial] [dev|debug|release|unit_tests] [device|simulator]"
+    echo -e "\tUSAGE: $0 [-c] [main|initial] [dev|debug|release|unit_tests] [device|simulator]"
```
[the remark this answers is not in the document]
ujjwal-cyph: changed

utilities/run_unit_tests.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 usage () {
     echo -e "\tUSAGE: $0 <main|initial> <device|simulator>"
```
kartikCypherock, line 4: Are we planning to change invocation of this script as well?

ujjwal-cyph: bash is more suitable. sh has some missing features.

ujjwal-cyph: for example, in `${OPTARG^}` the casing operation (`^`) is absent in `sh`.

kartikCypherock: I mean using -c/-f/-t flags wise

ujjwal-cyph: i misinterpreted. will update.

utilities/run_unit_tests.sh
```diff
@@ -1,55 +1,50 @@
...
+validate_platform
 
 echo "Invoking build script"
-echo "./utilities/build.sh "$1" "unit_tests" "$2""
-./utilities/build.sh "$1" "unit_tests" "$2"
+echo "./utilities/build.sh -c -f \"$FIRMWARE_TYPE\" -t \"unit_tests\" -p \"$BUILD_PLATFORM\""
+./utilities/build.sh -c -f "$FIRMWARE_TYPE" -t "unit_tests" -p "$BUILD_PLATFORM"
 
-if [ "$2" == "device" ]; then
+if [ "$BUILD_PLATFORM" == "device" ]; then
```
kartikCypherock, line 39: `BUILD_PLATFORM=${OPTARG^}` in line #27 converts "device" into "Device", therefore this comparison fails

kartikCypherock (approved):

PR #203, opened 2023-04-04
chore (git): Add git attributes

ujjwal-cyph:
Explicitly tell git to replace CRLF with LF


kartikCypherock (approved):

kartikCypherock:
Approving based on reference and tested it at my end

PR #207, opened 2023-04-05
Add code coverage for simulator build

TejasvCypherock:
This PR only consists of the coverage report for the simulator build, including gcov for device causes overflows, which needs to be further looked into.

- [x] Integrate code coverage results for simulator
- [ ] Integrate code coverage results for target device
- [ ] Add coverage to CI
- [ ] Exclude paths from coverage (eg. 3rd party libraries, official HAL sdk)


tests/sample_test/sample_test.c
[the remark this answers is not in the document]
TejasvCypherock: If the purpose was to demonstrate the sample tests, I think it need not stay any further as everyone in the team has now successfully created a test. Future developers can refer to existing new tests or the docs for the same.

utilities/run_unit_tests.sh
```diff
@@ -19,8 +19,6 @@ validate_platform () {
     fi
 }
 
-set -e
```
ujjwal-cyph, line 22: instead of deleting this line, we can also use ` || true` to skip or ignore unwanted errors

utilities/cmake/simulator/simulator.cmake
```diff
@@ -118,6 +118,10 @@ target_include_directories(${PROJECT_NAME} PRIVATE
         $<$<BOOL:UNIT_TESTS_SWITCH>:${PROJECT_SOURCE_DIR}/tests/sample_test>
```
kartikCypherock, line 118:
```suggestion
```

kartikCypherock: we should delete inclusion of this directory in include search paths as we have deleted the directory itself
same comment applicable to firmware.cmake

utilities/run_unit_tests.sh
```diff
@@ -19,8 +19,7 @@ validate_platform () {
...
-
+-set e
```
ujjwal-cyph, line 22: this is wrong

ujjwal-cyph:
Will be taking rest of the checklist tasks at a later stage.

ujjwal-cyph (approved):

ujjwal-cyph:
LGTM

PR #211, opened 2023-04-10
UI refactor with event getter

amanCypherock:
This change is part of https://app.clickup.com/t/9002019994/CHI-4708 ticket
- Integrates the UI events into UI screen callbacks
- Utilize UI events for correct operation in main event loop

The UI events usage will have to be updated further with the decided refactor of flows into apps.


rohanagarwal94:
Task linked: [CHI-4708 Implement UI event getter for firmware](https://app.clickup.com/t/9002019994/CHI-4708)

common/interfaces/user_interface/ui_input_text.c
```diff
@@ -498,11 +515,16 @@ static void next_btn_event_handler(lv_obj_t *next_btn, const lv_event_t event) {
...
     case LV_EVENT_CLICKED:
+      /* TODO: Update after refactor*/
+      /* Below statement sets the entered text to a global variable.
+         Correct action: A pointer should be provided by application to
+         fill the entered text to and it should be populated here, and
+         Dummy string should be replaced with user provided pointer */
       mark_input(data->entered_text);
       if (ui_mark_list_choice)
         (*ui_mark_list_choice)(data->current_index + 1);
+      ui_set_text_input_event("Dummy");
```
kartikCypherock, line 526: Let's return pointer to global variable (`flow_level.screen_input.input_text`), when the UI component is updated, we will just update the pointer from this point.

Application can rely on the `ui_event.text_ptr` in this case and will not need to update that

src/controller_main.c
```diff
@@ -105,8 +105,42 @@
 #include "pb_decode.h"
 #include "polygon.h"
 #include "rfc7539.h"
+#include "ui_events.h"
 #include "ui_instruction.h"
 
+ui_event_t os_ui_event;
```
kartikCypherock, line 111: I believe this is only required for testing flows now, so we should probably add a comment here so that we know that we need to remove this function after revamp is complete

common/interfaces/user_interface/ui_input_text.c
```diff
@@ -498,11 +515,16 @@ static void next_btn_event_handler(lv_obj_t *next_btn, const lv_event_t event) {
       }
       break;
     case LV_EVENT_CLICKED:
+      /* TODO: Update after refactor*/
+      /* Below statement sets the entered text to a global variable.
+         Correct action: A pointer should be provided by application to
+         fill the entered text to and it should be populated here, and
+         Dummy string should be replaced with user provided pointer */
       mark_input(data->entered_text);
       if (ui_mark_list_choice)
```
kartikCypherock, line 524: Are these 2 statements required?

common/interfaces/user_interface/ui_input_text.c
kartikCypherock: We can export one more API from UI event getter side to clear the input data for now.

common/interfaces/user_interface/ui_events.c
```diff
@@ -131,4 +134,13 @@ void ui_set_text_input_event(char *text_ptr) {
...
   ui_event.text_ptr = text_ptr;
-}
\ No newline at end of file
+  return;
+}
+
+void ui_fill_text(const char *text_src_ptr,
+                  char *input_text_ptr,
+                  const size_t max_text_len) {
+  ASSERT((text_src_ptr != NULL) && (input_text_ptr != NULL));
```
ujjwal-cyph: make this as check instead of ASSERT

common/interfaces/user_interface/ui_menu.c
```diff
@@ -58,9 +58,27 @@
...
+ *
+ * @return
+ * @retval
+ *
+ * @see
+ * @since v1.0.0
+ *
+ * @note
+ */
+void menu_create();
```
ujjwal-cyph: static?

address one is static

common/interfaces/user_interface/ui_list.c
```diff
@@ -61,9 +61,17 @@
...
+#include "ui_events_priv.h"
+
 static struct List_Data *data = NULL;
 static struct List_Object *obj = NULL;
 
+/**
+ * @brief Create UI for list
+ *
+ */
+void list_create();
```
ujjwal-cyph: static?

address one is made static.

src/main.c
```diff
@@ -199,8 +199,9 @@ int main(void) {
     /* Periodically call the lv_task handler.
      * It could be done in a timer interrupt or an OS task too.*/
     lv_task_handler();
+    process_ui_events();
     BSP_DelayMs(50);
-
+    process_ui_events();
```
ujjwal-cyph, line 204: duplicate function call

common/interfaces/user_interface/ui_list.c
kartikCypherock: fixed all instances

common/interfaces/user_interface/ui_events.c
kartikCypherock: fixed all instances

src/main.c
kartikCypherock: fixed

common/interfaces/user_interface/ui_input_text.c
ujjwal-cyph: resolving since the statements are removed

PR #215, opened 2023-04-14
Feat/nfc integrate os get evt

amanCypherock:
https://app.clickup.com/t/9002019994/CHI-4736

PR #214, opened 2023-04-14
Feat/nfc event/chi 4736

amanCypherock:
https://app.clickup.com/t/9002019994/CHI-4736

Added NFC events module to handle NFC card detect events.

PR #218, opened 2023-04-19
Integrate NFC events to OS event getter

amanCypherock:
https://app.clickup.com/t/9002019994/CHI-4736

Purpose of the PR is to add NFC events module  and integrate to the os event getter `get_events`

NOTE:- The changes in this PR are added above https://github.com/Cypherock/x1_wallet_firmware/pull/214 and https://github.com/Cypherock/x1_wallet_firmware/pull/215

With raising this PR both above PRs will be closed as those are now outdated.

Changelog from https://github.com/Cypherock/x1_wallet_firmware/pull/214 and https://github.com/Cypherock/x1_wallet_firmware/pull/215
- Broken the PN532 API for NFC A target detection to set command once and listen for a response when ready.
- Added card removal detection event to NFC events module.


common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+ */
+static bool check_card_removed_status(uint8_t retry_count);
+
+/**
+ * @brief   Handles response from read NFC A target init api
+ *          If card is selected, card select event is set, if an error is
+ * received, other than NFC_RESP_NOT_READY, state is updated to
+ * NFC_STATE_SET_SELECT_CARD_CMD
+ */
+static void nfc_handle_card_select_resp();
```
kartikCypherock, line 112: Let's add `void` as function parameter if the function does not expect any input. Applicable to all new functions.
Therefore, this function becomes: `static void nfc_handle_card_select_resp(void);`

For reference: https://app.clickup.com/9002019994/v/dc/8c8zv4u-20804/8c8zv4u-27384

common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+    nfc_state = NFC_STATE_WAIT_FOR_CARD_REMOVAL;
+    card_removal_retry_counter = 0;
+  }
+  return card_presence_state;
+}
+
+void nfc_task_handler() {
+  switch (nfc_state) {
+    case NFC_STATE_SET_SELECT_CARD_CMD:
+      if (pn532_set_nfca_target_init_command() == STM_SUCCESS)
```
kartikCypherock, line 187:
```suggestion
      if (pn532_set_nfca_target_init_command() == STM_SUCCESS) {
        nfc_state = NFC_STATE_WAIT_SELECT_CARD_RESP;
      }
```

common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+  if (card_presence_state == PN532_DIAGNOSE_CARD_DETECTED_RESP) {
+    nfc_state = NFC_STATE_WAIT_FOR_CARD_REMOVAL;
+    card_removal_retry_counter = 0;
+  }
+  return card_presence_state;
+}
+
+void nfc_task_handler() {
+  switch (nfc_state) {
+    case NFC_STATE_SET_SELECT_CARD_CMD:
```
kartikCypherock, line 186: Missing parenthesis around case : {}

common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+      if (pn532_set_nfca_target_init_command() == STM_SUCCESS)
+        nfc_state = NFC_STATE_WAIT_SELECT_CARD_RESP;
+      break;
+
+    case NFC_STATE_WAIT_SELECT_CARD_RESP: {
```
kartikCypherock, line 191: Can we please add `break;` statement inside case parenthesis?

For example:
```
case ABC: {
  break;
}
```

tests/nfc/nfc_event_getter_test.c
```diff
@@ -0,0 +1,55 @@
...
+
+  instruction_scr_init("Place a card", "NFC EVENT TEST");
+  lv_task_handler();
+
+  nfc_en_select_card_task();
+
+  get_events(evt_config, &evt_status);
+
+  if (evt_status.nfc_event.event_occured) {
+    instruction_scr_change_text("Card Detected", true);
```
kartikCypherock, line 26: Further check the `event_type`: it should be either `DETECTED` or `REMOVED`

common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+  nfc_event.event_type = NFC_EVENT_CARD_REMOVED;
+  nfc_state = NFC_STATE_CARD_REMOVED;
+}
+
+void nfc_en_select_card_task() {
+  nfc_state = NFC_STATE_SET_SELECT_CARD_CMD;
+}
+
+uint32_t nfc_en_wait_for_card_removal_task() {
+  uint32_t card_presence_state = nfc_diagnose_card_presence();
```
kartikCypherock, line 176: 1. Why is there a requirement to check for card presence in this case? 
-> If we need it, then for consistency, we should also check if card is **not** present when `nfc_en_select_card_task()` is called

tests/nfc/nfc_event_getter_test.c
```diff
@@ -0,0 +1,55 @@
...
+  } else {
+    instruction_scr_change_text("Something bad happened", true);
+  }
+
+  nfc_select_card();
+  instruction_scr_change_text("Remove card", true);
+
+  uint32_t err = nfc_en_wait_for_card_removal_task();
+
+  if (err != 0) {
```
kartikCypherock, line 37:
```suggestion
  if (err != SUCCESS_) {
```

tests/nfc/nfc_event_getter_test.c
```diff
@@ -0,0 +1,55 @@
...
+    LOG_CRITICAL("xxx39 %08x", err);
+    instruction_scr_change_text("Something bad happened 2", true);
+  } else {
+    get_events(evt_config, &evt_status);
+
+    if (evt_status.nfc_event.event_occured) {
+      instruction_scr_change_text("Card Removed", true);
```
kartikCypherock, line 44: Further check the `event_type`: it should be either `DETECTED` or `REMOVED`

tests/nfc/nfc_event_getter_test.c
```diff
@@ -0,0 +1,55 @@
+#include "board.h"
+#include "buzzer.h"
+#include "events.h"
+#include "nfc.h"
+#include "nfc_events.h"
+#include "ui_instruction.h"
+
+#ifdef NFC_EVENT_CARD_DETECT_MANUAL_TEST
+
+void test_nfc_operation_with_os_getter() {
```
kartikCypherock, line 10: Let's add this test in `unit_tests_list.c` under the same compilation macro `NFC_EVENT_CARD_DETECT_MANUAL_TEST`. We should add a brief comment there as to why that test is added in a special way.

common/interfaces/card_interface/nfc_events.c
```diff
@@ -0,0 +1,221 @@
...
+    } break;
+
+    default:
+      LOG_ERROR("xxx37: %d", nfc_state);
+      break;
+  }
+}
+
+void nfc_ctx_destroy() {
+  if (nfc_state != NFC_STATE_OFF) {
```
kartikCypherock, line 216: This check seems to be redundant as it's guaranteed that `nfc_ctx_destroy` is called once per call to `get_event`

kartikCypherock (changes requested):

tests/nfc/nfc_event_getter_test.c
amanCypherock: good catch

tests/nfc/nfc_event_getter_test.c
amanCypherock: Actually this should be compared with PN532_DIAGNOSE_CARD_DETECTED_RESP as we are not returning an internal success code, but status returned from diagnose command

tests/nfc/nfc_event_getter_test.c
amanCypherock: I feel this could be called in unit_test_main.c rather than unit_test_list.c as the test is of manual category, otherwise another test could be waiting for this test to complete when enabled.

common/interfaces/card_interface/nfc_events.c
amanCypherock: not per call to `get_event`, only when nfc event is enabled in `get_event`.
Will remove this check though.

common/interfaces/card_interface/nfc_events.c
amanCypherock: Although none of the cases need a parenthesis as all variable declarations are made in static function, is there any specific reason to keep the `break` statement inside the parenthesis?

common/interfaces/card_interface/nfc_events.c
amanCypherock: We don't really need to add one here, do we?

common/interfaces/card_interface/nfc_events.c
amanCypherock: Will add the change manually, suggested change will add line 188 without removing it.

common/interfaces/card_interface/nfc_events.c
amanCypherock: As though why the check was introduced, if PN532 is not in the correct state (where a card is already selected/activated) diagnose command would return error even if a card was present in proximity, which leads to false result from the event call.
This event is only applicable when waiting for removal of an already activated card.

Such is not the case with `nfc_en_select_card_task()`, hence a check is not required there.

rohanagarwal94:
Task linked: [CHI-4736 Implement NFC Event getter](https://app.clickup.com/t/9002019994/CHI-4736)

common/interfaces/card_interface/nfc_events.c
kartikCypherock: Please add a comment in code explaining why this check is needed.

tests/nfc/nfc_event_getter_test.c
```diff
@@ -0,0 +1,55 @@
...
+
+    if (evt_status.nfc_event.event_occured) {
+      instruction_scr_change_text("Card Removed", true);
+      buzzer_start(500);
+    } else {
+      instruction_scr_change_text("Something bad happened 3", true);
+    }
+  }
+
+  BSP_DelayMs(5000);
```
kartikCypherock, line 51: I think there should be a test which tests the use case of p0 event.

Assume application enabled card detect event but a p0 event occurred... will the application be able to detect a card detect event at a later point?

tests/nfc/nfc_event_getter_test.c
kartikCypherock: No, we should not create exception in the way we are adding tests.
We should create this test in the same way we create other tests, only keep it under compilation macro guard.

So, please create a test group etc, as done in other modules.

common/interfaces/card_interface/nfc_events.c
amanCypherock: updated

common/interfaces/card_interface/nfc_events.c
amanCypherock: updated

common/interfaces/card_interface/nfc_events.c
amanCypherock: added

common/interfaces/card_interface/nfc_events.c
amanCypherock: added

common/interfaces/card_interface/nfc_events.h
```diff
@@ -0,0 +1,95 @@
...
+  NFC_EVENT_CARD_DETECT = 1,
+  NFC_EVENT_CARD_REMOVED
+} nfc_event_type_t;
+
+typedef struct {
+  bool event_occured;
+  nfc_event_type_t event_type;
+} nfc_event_t;
+
+typedef enum {
```
ujjwal-cyph, line 35: can move this to `_priv` header

tests/nfc/nfc_event_getter_test.c
amanCypherock: I agree that this type of test is not complete as this does not handle all possible cases. The test was introduced to be able to verify integration with `get_events` but if all cases are handled it becomes almost a flow.

I suggest we add that change in a separate PR

tests/nfc/nfc_event_getter_test.c
amanCypherock: added

tests/nfc/nfc_event_getter_test.c
amanCypherock: added

tests/nfc/nfc_event_getter_test.c
amanCypherock: added

common/interfaces/card_interface/nfc_events.h
amanCypherock: No need there either, moving to .c file

stm32-hal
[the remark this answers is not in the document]
amanCypherock: added

common/interfaces/card_interface/nfc_events.c
kartikCypherock: I think the comment was not pushed as I cannot see it in the diff

kartikCypherock: Added in header file so resolving this comment

stm32-hal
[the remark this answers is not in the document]
ujjwal-cyph: https://github.com/Cypherock/x1_wallet_hal_stm32/pull/5

kartikCypherock (approved):

amanCypherock:
Need to merge https://github.com/Cypherock/x1_wallet_hal_stm32/pull/5 first

common/core/events.c
```diff
@@ -134,5 +135,10 @@ void get_events(evt_config_t evt_config, evt_status_t *p_evt_status) {
 
   /* Any post cleanup required */
   p0_ctx_destroy();
+
+  if (evt_config.evt_selection.bits.nfc_events) {
+    nfc_ctx_destroy();
```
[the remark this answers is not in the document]
ujjwal-cyph: empty function is not needed :)

PR #222, opened 2023-04-27
feat (build): Integrate protobuf using nanopb

ujjwal-cyph:
Integrate protobuf using nanopb into build system.

Fixes https://app.clickup.com/t/9002019994/CHI-3948

Due to a limitation of nanopb scripts (Ref: https://github.com/nanopb/nanopb/issues/527), it not possible to direct generated protob sources to desired directory via cmake configuration. So we cannot use cmake configuration of nanopb and a bash script is being used inside cmake.
We can look into this at a later stage.


rohanagarwal94:
Task linked: [CHI-3948 Protobuf build integration for SDK](https://app.clickup.com/t/9002019994/CHI-3948)

utilities/cmake/simulator/simulator.cmake
```diff
@@ -38,6 +38,9 @@ endif()
 target_include_directories(${PROJECT_NAME} PRIVATE
         src/
 
+        ${CMAKE_CURRENT_BINARY_DIR}
```
kartikCypherock, line 41: What is the need to add `${CMAKE_CURRENT_BINARY_DIR)` into the include path?

kartikCypherock:
I have one generic comment related to file deletions in `common/libraries/nanopb`, my understanding is that these files were required to support wallet connect related communication. I hope there is no impact expected on current features if we use generated pb decoders instead of a tested decoder file?

utilities/cmake/simulator/simulator.cmake
ujjwal-cyph: the generated files are at the root of cmake build directory hence need to provide that for including headers.

ujjwal-cyph:
> I have one generic comment related to file deletions in `common/libraries/nanopb`, my understanding is that these files were required to support wallet connect related communication. I hope there is no impact expected on current features if we use generated pb decoders instead of a tested decoder file?

Did a diff comparison of the files. Following differences noticed:
- nanopb `v0.4.6` is being updated to `v0.4.7` with this PR
  - no concerns IMO
- nanopb was being used with malloc
  - corrected this (added macro switch in the compile definition)

utilities/cmake/simulator/simulator.cmake
ujjwal-cyph: i tried `src/../` in place of `${CMAKE_CURRENT_BINARY_DIR}` but did not work. The generated protobuf headers could not be located.

kartikCypherock:
> > I have one generic comment related to file deletions in `common/libraries/nanopb`, my understanding is that these files were required to support wallet connect related communication. I hope there is no impact expected on current features if we use generated pb decoders instead of a tested decoder file?
> 
> Did a diff comparison of the files. Following differences noticed:
> 
> * nanopb `v0.4.6` is being updated to `v0.4.7` with this PR
>   
>   * no concerns IMO
> * nanopb was being used with malloc
>   
>   * corrected this (added macro switch in the compile definition)

Ok, cool
Let's loop in @TejasvCypherock to confirm all compilation definitions set in this PR for nanopb match the same as done before

utilities/cmake/simulator/simulator.cmake
ujjwal-cyph: The example provided by nanopb (ref `vendor/nanopb/extra/FindNanopb.cmake#63`) mentions to use `CMAKE_CURRENT_BINARY_DIR` cmake variable for including protoc generated headers.

utilities/cmake/simulator/simulator.cmake
```diff
@@ -20,7 +20,7 @@ set(EXECUTABLE ${PROJECT_NAME})
 find_package(SDL2 REQUIRED SDL2)
 include_directories(${SDL2_INCLUDE_DIRS})
 
-add_executable(${EXECUTABLE} ${SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/version.c ${INCLUDES})
+add_executable(${EXECUTABLE} ${SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/version.c ${COMMON_SRCS} ${COMMON_HDRS} ${PROTO_SRCS} ${PROTO_HDRS} ${INCLUDES})
```
kartikCypherock, line 23:
```suggestion
add_executable(${EXECUTABLE} ${SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/version.c ${PROTO_SRCS} ${PROTO_HDRS} ${INCLUDES})
```

kartikCypherock: We don't need `$COMMON_SRCS` and `$COMMON_HDRS` anymore

CMakeLists.txt
```diff
@@ -31,4 +52,12 @@ if("${CMAKE_BUILD_PLATFORM}" STREQUAL "Device")
     include(utilities/cmake/firmware/firmware.cmake)
 else()
     include(utilities/cmake/simulator/simulator.cmake)
-endif()
\ No newline at end of file
+endif()
+
+# Include nanopb (at ${NANOPB_INCLUDE_DIRS}) and protoc-generated (at
+# ${CMAKE_CURRENT_BINARY_DIR}) headers
```
ujjwal-cyph, line 58:
```suggestion
# ${CMAKE_CURRENT_BINARY_DIR}) headers. Ref: vendor/nanopb/extra/FindNanopb.cmake#63
```

amanCypherock:
We should also check all CI-related changes which are not triggered already.

ujjwal-cyph:
> We should also test all CI-related changes which are not triggered already.

Tested and fixed.

Reference run: https://github.com/ujjwal-cyph/x1_wallet_firmware/actions/runs/4828683503

TejasvCypherock (changes requested):
Need to remove old nanopb path from include in firmware.cmake

TejasvCypherock (commented):
We could also rename the `protob` submodule to its original name i.e. `cypherock-common`. As we might want to add other files that are not related to protobuf

CMakeLists.txt
```diff
@@ -15,6 +15,27 @@ endif()
...
+file(GLOB_RECURSE PROTO_SRC common/protob/proto/*.proto)
+
+# --generated-include-format specifies the include format. Two viable options
+# are "#include <%s>" and "#include \"proto/%s\"".
+# The default include format "#include \"%s\"" creates problems in cases of
+# duplicate file names residing under different packages (eg. error.proto and
+# btc.error.proto) as it would result into a collision.
+# By using <> format instead of "proto/" helps because in future, relocation
+# of generated files would not require updating this parameter.
+set(NANOPB_OPTIONS ",--generated-include-format=#include <%s>")
```
TejasvCypherock, line 32: We might also want to consider adding a `--c-style` flag to better match our existing coding style

TejasvCypherock (changes requested):

CMakeLists.txt
ujjwal-cyph: added

TejasvCypherock (changes requested):
As the files are directly generated in the build folder, It is not clear to me how will we be using these files while programming as the IDE would need its reference for intellisense and other features. You might want to set up a simple example to clarify this situation. Also the `--c-style` flag doesn't seem to work.

kartikCypherock (approved):

ujjwal-cyph:
> As the files are directly generated in the build folder, It is not clear to me how will we be using these files while programming as the IDE would need its reference for intellisense and other features. You might want to set up a simple example to clarify this situation. Also the `--c-style` flag doesn't seem to work.

- Fixed the issue with `--c-style` parameter (https://github.com/Cypherock/x1_wallet_firmware/pull/222/commits/4eb09cc109839b0e462e03c0c84495e065132653).
- Added a bash script to build protob sources for IDE support (https://github.com/Cypherock/x1_wallet_firmware/pull/222/commits/e7704807f1e6bff751562346b12f34c9f3d05827).

TejasvCypherock (approved):

PR #223, opened 2023-04-28
feat(core): Implement flow engine for event handling and ux navigation

kartikCypherock:
Fixes ticket: https://app.clickup.com/t/9002019994/CHI-4820


common/core/flow_engine.c
```diff
@@ -0,0 +1,265 @@
...
+/*****************************************************************************
+ * STATIC VARIABLES
+ *****************************************************************************/
+static flow_step_t *engine_buffer_0[ENGINE_STACK_DEPTH] = {0};
+static array_list_t engine_list_0;
+
+static flow_step_t *engine_buffer_1[ENGINE_STACK_DEPTH] = {0};
+static array_list_t engine_list_1;
+
+static engine_ctx_t engine_ctx[ENGINE_NUM_BUFFERS] = {
```
amanCypherock, line 106: We should add an API to return the index of the next free buffers. In cases where more than 2 buffers are required, use of an incorrect or hardcoded index can potentially alter the state of the device when returning from the application.

common/core/flow_engine.c
```diff
@@ -0,0 +1,265 @@
...
+
+    /* Break if the stack or queue reached an empty state */
+    if ((false == engine_get_current_flow_step(engine_buffer, &current_flow)) ||
+        (NULL == current_flow)) {
+      break;
+    }
+
+    const evt_config_t *evt_config_ptr = current_flow->evt_cfg_ptr;
+    /* We must ensure that the `evt_config_ptr` is not NULL */
+    ASSERT(NULL != evt_config_ptr);
```
amanCypherock, line 230: EVT pointer being null should be checked on adding the pointer to the list in `engine_add_next_flow_step`

amanCypherock (changes requested):

common/core/flow_engine.c
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  /* We want to store the pointer to the flow step inputted from the argument
+   * and NOT the actual content, therefore we need to pass the double pointer to
+   * the API */
+  const flow_step_t **flow_step_dptr = &flow_step_ptr;
+  result = array_list_insert(engine_ctx[engine_buffer].array_list_config,
+                             flow_step_dptr);
+  return result;
+}
```
amanCypherock, line 166: Should check here if `ENGINE_STACK_DEPTH` has been reached. A buffer overflow could be overlooked currently.

amanCypherock:
Flow engine tests need to be updated

common/core/flow_engine.c
```diff
@@ -0,0 +1,265 @@
...
+ *****************************************************************************/
+static flow_step_t *engine_buffer_0[ENGINE_STACK_DEPTH] = {0};
+static array_list_t engine_list_0;
+
+static flow_step_t *engine_buffer_1[ENGINE_STACK_DEPTH] = {0};
+static array_list_t engine_list_1;
+
+static engine_ctx_t engine_ctx[ENGINE_NUM_BUFFERS] = {
+    {&engine_buffer_0[0], &engine_list_0},
+    {&engine_buffer_1[0], &engine_list_1}};
```
amanCypherock, line 108: Since, the context of an app or flow is maintained by the object of `array_list` and buffer, making it local for the `flow_engine` restricts the usage of `array_list` and relies on an index to control the objects from. This leaves room for incorrect buffer usage when hard-coded indexes are used.

An alternative is to let an application declare its own buffers with its own length etc. The buffers can be passed to the relevant callbacks to manipulate the flow.

rohanagarwal94:
Task linked: [CHI-4820 X1 vault task framework](https://app.clickup.com/t/9002019994/CHI-4820)

common/libraries/util/array_list.c
```diff
@@ -0,0 +1,244 @@
...
+/*****************************************************************************
+ * GLOBAL FUNCTIONS
+ *****************************************************************************/
+bool array_list_initialize(array_list_t *const list,
+                           void *array_buffer,
+                           size_t size_of_element,
+                           size_t capacity) {
+  /* Must ensure that the array_buffer is not NULL and the size_of_element is
+   * non-zero */
+  if ((NULL == list) || (NULL == array_buffer) || (0 == size_of_element)) {
```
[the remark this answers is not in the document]
kartikCypherock: api not required anymore

common/libraries/util/array_list.c
```diff
@@ -0,0 +1,244 @@
...
+
+  memset(list, 0, sizeof(array_list_t));
+
+  list->array = array_buffer;
+  list->size_of_element = size_of_element;
+  list->current_index = 0;
+  list->num_of_elements = 0;
+  list->max_capacity = capacity / (list->size_of_element);
+
+  return;
```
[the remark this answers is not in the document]
kartikCypherock: api not required anymore

common/core/flow_engine.c
kartikCypherock: updated engine to accept buffers from external world, so the user of the engine needs to keep track of this now

common/core/flow_engine.c
kartikCypherock: added this check.. good catch

common/core/flow_engine.c
kartikCypherock: I have updated the engine APIs to accept buffer from flows and the context is passed on to the callbacks as well..

src/main.c
```diff
@@ -173,7 +173,7 @@ int main(void) {
     }
 
     if (sys_flow_cntrl_u.bits.nfc_off == false) {
-      nfc_deselect_card();
+      // nfc_deselect_card();
```
ujjwal-cyph, line 176: probably not needed to comment this out ATM

kartikCypherock: fixed

common/core/flow_engine.h
```diff
@@ -0,0 +1,184 @@
...
+ * @param ctx Pointer to data of type engine_ctx_t which holds the correct data
+ * for the buffer
+ * @param flow_step_dptr A double pointer which will be filled with the element
+ * of type flow_step_t*, which can be dereferenced by the caller.
+ * @return true If the element was returned successfully
+ * @return false If the element was not returned: It could be due to incorrect
+ * parameters, or because the buffer is EMPTY.
+ */
+bool engine_get_current_flow_step(engine_ctx_t *ctx,
+                                  flow_step_t **flow_step_dptr);
```
amanCypherock, line 161: Do we need this as a global API?

kartikCypherock: good point, we can keep it static

kartikCypherock: i've fixed it

amanCypherock (approved):

ujjwal-cyph (approved):

PR #230, opened 2023-05-06
refactor (usb): Restructure code organization

ujjwal-cyph:


common/interfaces/desktop_app_interface/usb_api_priv.h
```diff
@@ -16,22 +16,189 @@
...
+} comm_status_t;
+
 /*****************************************************************************
  * EXPORTED VARIABLES
  *****************************************************************************/
 
+// received/ready-to-send cmd can be decoded/encoded into this buffer
+extern uint8_t comm_io_buffer[COMM_BUFFER_SIZE];
+extern comm_payload_t comm_payload;
+extern comm_status_t comm_status;
```
amanCypherock, line 159: Preferred to expose these variables through APIs in this file. This is according to the defined coding guidelines.

common/interfaces/desktop_app_interface/usb_api_priv.h
```diff
@@ -16,22 +16,189 @@
...
+
 /*****************************************************************************
  * GLOBAL FUNCTION PROTOTYPES
  *****************************************************************************/
+
+static inline void comm_set_payload_struct(uint16_t proto_len,
```
amanCypherock, line 165: ATCG: Inline functions are not preferred as they are harder to mock for unit tests. Suggest moving the function definition to source file.

common/interfaces/desktop_app_interface/usb_rx_parser.c
```diff
@@ -0,0 +1,296 @@
+/**
+ * @file    usb_rx_parser.c
```
kartikCypherock, line 2: consider renaming to `comm_rx_parser.c` as it is independent of physical layer

common/interfaces/desktop_app_interface/usb_internals.c
```diff
@@ -0,0 +1,479 @@
+/**
+ * @file    usb_internal.c
```
kartikCypherock, line 2: consider renaming to `comm_internal.c` as it is independent of physical layer

common/interfaces/desktop_app_interface/usb_rx_parser.c
kartikCypherock: not valid because we are calling lusb functions, therefore this comment is redundant

kartikCypherock (approved):

common/interfaces/desktop_app_interface/usb_api_priv.h
ujjwal-cyph: will take this in next change request

amanCypherock: Resolving as extern declared only in _priv header. But for future changes, we should not allow such usage, deviating from coding guidelines.

common/interfaces/desktop_app_interface/usb_api_priv.h
amanCypherock: Resolving as not introduced in this pr and defined in _priv file

amanCypherock (approved):

common/interfaces/desktop_app_interface/usb_api_priv.h
amanCypherock: Preferred to change this PR as introduced here only

ujjwal-cyph: chnaged

PR #232, opened 2023-05-08
refactor: Use protobuf struct for status payload

ujjwal-cyph:
Fixes https://app.clickup.com/t/9002019994/CHI-4826

### DIFF
--- a/common/interfaces/desktop_app_interface/usb_internals.c
+++ b/common/interfaces/desktop_app_interface/usb_internals.c
@@ -73,6 +73,7 @@
 #include "logger.h"
 #include "p0_events.h"
 #include "pb_encode.h"
+#include "status_api.h"
 #include "sys_state.h"
 #include "usb_api_priv.h"
 #include "utils.h"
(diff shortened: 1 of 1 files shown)


rohanagarwal94:
Task linked: [CHI-4826 Protobuf integration into core](https://app.clickup.com/t/9002019994/CHI-4826)

common/interfaces/desktop_app_interface/usb_internals.c
```diff
@@ -303,25 +304,31 @@ static comm_error_code_t comm_process_abort_packet(const packet_t *rx_packet) {
...
+  payload[3] = 0x00;    // dummy raw length
+
+  // reserve space for length of streams
+  uint8_t len_size = 4;
+  core_status_t status = get_core_status();
+  pb_ostream_t stream =
+      pb_ostream_from_buffer(payload + len_size, sizeof(payload) - len_size);
+
+  status.current_cmd_seq = comm_status.curr_cmd_seq_no;
+  status.cmd_state = comm_status.curr_cmd_state;
```
kartikCypherock, line 319: Can we change type of `curr_cmd_state` to `core_cmd_state_t`

common/interfaces/desktop_app_interface/usb_internals.c
```diff
@@ -303,25 +304,31 @@ static comm_error_code_t comm_process_abort_packet(const packet_t *rx_packet) {
  */
 static void send_status_packet(const packet_t *rx_packet) {
   uint8_t payload[COMM_MAX_PAYLOAD_SIZE];
```
kartikCypherock, line 306: Since we are only sending one chunk out of one total chunk (call to `comm_write_packet`), therefore, we are fixing the length of data to be sent via this function.

Consider a case, wherein protobuf encoding passed due to the `6KiB` buffer size, but it will not be easily detected since the physical layer will only transfer 1 packet of data. 

So consider reducing size of `payload` with appropriate size

common/interfaces/desktop_app_interface/usb_internals.c
```diff
@@ -303,25 +304,31 @@ static comm_error_code_t comm_process_abort_packet(const packet_t *rx_packet) {
...
+  payload[2] = 0x00;
+  payload[3] = 0x00;    // dummy raw length
+
+  // reserve space for length of streams
+  uint8_t len_size = 4;
+  core_status_t status = get_core_status();
+  pb_ostream_t stream =
+      pb_ostream_from_buffer(payload + len_size, sizeof(payload) - len_size);
+
+  status.current_cmd_seq = comm_status.curr_cmd_seq_no;
```
kartikCypherock, line 318: Please add a comment as to why we are updating fields of variable of type `core_status_t` to make our intentions clear

common/interfaces/desktop_app_interface/usb_internals.c
kartikCypherock: This was my wrong understanding regarding `COMM_MAX_PAYLOAD_SIZE`. I got confused and thought this was a 6KiB buffer size.

This comment is not valid anymore, so closing the conversation.

ujjwal-cyph: the size of `payload` is 48.

it is not  `6KiB`

common/interfaces/desktop_app_interface/usb_internals.c
kartikCypherock: Also, let's plan to retire members of `comm_status` which should not be controlled by this layer

amanCypherock (approved):

common/interfaces/desktop_app_interface/usb_internals.c
ujjwal-cyph: yes. taking this up as separate change request.

PR #234, opened 2023-05-09
Fix/webusb collision

amanCypherock:


common/core/status_api.c
```diff
@@ -95,6 +97,17 @@ core_status_t core_status = CORE_STATUS_INIT_ZERO;
...
   core_status.device_idle_state = idle_state;
+
+  /**
+   * If app state is idle, active interface is cleared to allow recieving
+   * commands from any interface. If app state is not idle, it means an
+   * application/flow is already in progress and command from a new interface
+   * shouldn't be allowed before tasks of the app have been completed or app
+   * is closed.
+   */
+  if (core_status.device_idle_state ==
```
kartikCypherock, line 108: ATCG: 
```suggestion
  if (CORE_DEVICE_IDLE_STATE_DEVICE_IDLE_STATE_IDLE == core_status.device_idle_state)
```

common/interfaces/desktop_app_interface/usb_internals.c
```diff
@@ -407,6 +414,10 @@ static void comm_write_packet(const uint16_t chunk_number,
 /*****************************************************************************
  * GLOBAL FUNCTIONS
  *****************************************************************************/
+void comm_reset_interface(void) {
```
kartikCypherock, line 417: Can we move this function to usb_api.c?

amanCypherock: I've defined this function in the file where the variable is also declared. What's the motivation to move this?

common/interfaces/desktop_app_interface/usb_rx_parser.c
```diff
@@ -119,34 +119,38 @@ static uint8_t SDK_REQ_PACKET[] = {COMM_V0_START_OF_HEADER,
...
-                                    0x00,
-                                    0x03,
-                                    0x00,
-                                    0x06,
-                                    0x00,
-                                    0x00,
-                                    0x0F,
-                                    0x30};
+static uint8_t SDK_RESP_PACKET[] = {
+    COMM_V0_START_OF_HEADER,
```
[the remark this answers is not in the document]
amanCypherock: This is to keep support for the current SDK version. When we actually depreciate it, we won't need this one anyway.

common/core/status_api.c
amanCypherock: updated

common/interfaces/desktop_app_interface/usb_internals.c
kartikCypherock: Motivation is simply to keep all APIs in api file and not internal file

kartikCypherock (approved):

PR #237, opened 2023-05-10
feat (build): Add bash script to auto-generate options file

kartikCypherock:
To integrate some protobuf structures in our project, we need to configure options for the nanopb compiler. 
This PR generates the options files for each proto file and configures nanopb compiler to use these options files

Of course, these options file will be modified later as per usage.

### DIFF
--- a/utilities/proto/gen-options.sh
+++ b/utilities/proto/gen-options.sh
@@ -10,10 +10,17 @@ function wrong_dir {
 test -d common/cypherock-common/proto || wrong_dir
 
 # Make sure all directories exist for the proto-options
-test -d common/proto-options/ || mkdir common/proto-options/
-test -d common/proto-options/manager || mkdir common/proto-options/manager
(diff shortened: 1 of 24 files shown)


utilities/proto/gen-options.sh
```diff
@@ -0,0 +1,58 @@
...
+  exit 1
+}
+
+test -d common/cypherock-common/proto || wrong_dir
+
+# Make sure all directories exist for the proto-options
+test -d common/proto-options/ || mkdir common/proto-options/
+test -d common/proto-options/manager || mkdir common/proto-options/manager
+test -d common/proto-options/btc || mkdir common/proto-options/btc
+test -d common/proto-options/evm || mkdir common/proto-options/evm
```
ujjwal-cyph, line 16: make this dynamic. so new folder addition will not break the script

kartikCypherock: fixed

ujjwal-cyph (approved):

PR #245, opened 2023-05-13
feat: Base structure for usb callback to pass control to applications

kartikCypherock:
This implements a very basic structure for the USB callback to enable invocation of applications from the host interface


amanCypherock (approved):

amanCypherock:
Note: The folder structure needs to be updated for all apps

PR #250, opened 2023-05-15
feat: Add function to populate device info fields

ujjwal-cyph:
Fixes https://app.clickup.com/t/9002019994/PRF-4737

### DIFF
--- a/apps/manager_app/get_device_info.c
+++ b/apps/manager_app/get_device_info.c
@@ -1,7 +1,7 @@
 /**
- * @file    manager_app.c
+ * @file    get_device_info.c
  * @author  Cypherock X1 Team
- * @brief
+ * @brief   Populates device info fields at runtime requests.
  * @copyright Copyright (c) 2023 HODL TECH PTE LTD
  * <br/> You may obtain a copy of license at <a href="https://mitcc.org/"
  *target=_blank>https://mitcc.org/</a>
@@ -59,10 +59,10 @@
 /*****************************************************************************
  * INCLUDES
  *****************************************************************************/
-#include "manager_app.h"
 
-#include "manager_api.h"
-#include "onboarding.h"
+#include "controller_level_four.h"
+#include "flash_api.h"
+#include "manager_app.h"
 
 /*****************************************************************************
  * EXTERN VARIABLES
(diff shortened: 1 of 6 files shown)


apps/manager_app/get_device_info.c
```diff
@@ -0,0 +1,150 @@
...
+ * @param firmware_version
+ * @return bool Status indicating the
+ */
+bool get_firmware_version(common_version_t *firmware_version);
+
+/*****************************************************************************
+ * STATIC FUNCTIONS
+ *****************************************************************************/
+
+bool get_firmware_version(common_version_t *firmware_version) {
```
kartikCypherock, line 103: How about we refactor the `get_fwVer()` API directly to return `common_version_t` instead of making more and more wrapper functions?

All other calls needs to be updated anyway

Just a suggestion, we can choose to keep it this way only

rohanagarwal94:
Task linked: [PRF-4737 Host connection detection](https://app.clickup.com/t/9002019994/PRF-4737)

apps/manager_app/get_device_info.c
amanCypherock: If this was a custom structure, then it would have been okay, but `common_version_t` is a nanopb-generated structure, when including this into flash_api, we would be creating unnecessary dependency in that file.

If this were a custom-defined structure, then the suggestion is valid.

apps/manager_app/get_device_info.c
```diff
@@ -0,0 +1,150 @@
...
+  firmware_version->minor = (version >> 16) & 0xFF;
+  firmware_version->patch = version & 0xFFFF;
+  return true;
+}
+
+/*****************************************************************************
+ * GLOBAL FUNCTIONS
+ *****************************************************************************/
+
+manager_get_device_info_response_t get_device_info() {
```
kartikCypherock, line 120:
```suggestion
manager_get_device_info_response_t get_device_info(void) {
```

apps/manager_app/manager_app.c
```diff
@@ -105,6 +106,17 @@ void manager_app_main(usb_event_t usb_evt) {
...
+      uint8_t response[MANAGER_GET_DEVICE_INFO_RESULT_RESPONSE_SIZE] = {0};
+      manager_result_t result = MANAGER_RESULT_INIT_ZERO;
+      result.which_response = MANAGER_RESULT_GET_DEVICE_INFO_TAG;
+      result.response.get_device_info = get_device_info();
+      ASSERT(encode_manager_result(&result,
+                                   response,
+                                   MANAGER_GET_DEVICE_INFO_RESULT_RESPONSE_SIZE,
+                                   &msg_size));
+      usb_send_msg(response, msg_size);
+      onboarding_set_static_screen();
```
kartikCypherock, line 119: Does it require any condition in future?
If yes, please add a Todo

apps/manager_app/get_device_info.c
kartikCypherock: By the way as per our last discussion, we decided to check the query type in every step of the flow, so do you think this should also be done here before the operation starts?

apps/manager_app/manager_app.c
ujjwal-cyph: added

apps/manager_app/get_device_info.c
ujjwal-cyph: moved this logic and added the necessary checks

apps/manager_app/manager_app.h
```diff
@@ -46,4 +47,16 @@
...
+ * @brief Return a filled instance of get_device_info_response_t.
+ */
+manager_get_device_info_response_t get_device_info(void);
+
+/**
+ * @brief Initiates the get device info flow.
+ *
+ * @param query Reference to the decoded query struct from the host app
+ */
+void get_device_info_flow(const manager_query_t *query);
```
amanCypherock, line 60: Let's move these functions to `manager_app_priv.h`
We should not expose the individual functions/flows of an app.

amanCypherock: refer https://github.com/Cypherock/x1_wallet_firmware/compare/refactor/core-framework...test/support-protobuf-auth-card?expand=1#diff-20226d7b084854f33b609e942be568155e853af7c07d575547acb1c125824ceb

apps/manager_app/get_device_info.c
```diff
@@ -0,0 +1,185 @@
...
+ * @param firmware_version
+ * @return bool Status indicating the
+ */
+bool get_firmware_version(common_version_t *firmware_version);
+
+/*****************************************************************************
+ * STATIC FUNCTIONS
+ *****************************************************************************/
+
+bool get_firmware_version(common_version_t *firmware_version) {
```
amanCypherock, line 105: add `static`

kartikCypherock (approved):

apps/manager_app/get_device_info.c
```diff
@@ -0,0 +1,185 @@
...
+    result.which_response = MANAGER_RESULT_GET_DEVICE_INFO_TAG;
+    result.response.get_device_info.which_response =
+        MANAGER_GET_DEVICE_INFO_RESPONSE_CORE_ERROR_TAG;
+    result.response.get_device_info.response.core_error.which_error =
+        ERROR_CORE_ERROR_UNKNOWN_ERROR_TAG;
+    result.response.get_device_info.response.core_error.error.unknown_error = 1;
+    encode_manager_result(&result,
+                          response,
+                          MANAGER_GET_DEVICE_INFO_RESULT_RESPONSE_SIZE,
+                          &msg_size);
```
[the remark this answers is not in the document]
amanCypherock: Should also check encoding result

apps/manager_app/get_device_info.c
```diff
@@ -0,0 +1,185 @@
...
+
+  if (MANAGER_QUERY_GET_DEVICE_INFO_TAG != query->which_request ||
+      MANAGER_GET_DEVICE_INFO_REQUEST_INITIATE_TAG !=
+          query->request.get_device_info.which_request) {
+    result.which_response = MANAGER_RESULT_GET_DEVICE_INFO_TAG;
+    result.response.get_device_info.which_response =
+        MANAGER_GET_DEVICE_INFO_RESPONSE_CORE_ERROR_TAG;
+    result.response.get_device_info.response.core_error.which_error =
+        ERROR_CORE_ERROR_UNKNOWN_ERROR_TAG;
+    result.response.get_device_info.response.core_error.error.unknown_error = 1;
```
amanCypherock, line 167: Seems this part is common in the above API as well, can we create a separate function for this?

apps/manager_app/get_device_info.c
ujjwal-cyph: added

apps/manager_app/manager_app.h
```diff
@@ -46,4 +47,16 @@
  */
 void manager_app_main(usb_event_t usb_evt);
 
+/**
+ * @brief Return a filled instance of get_device_info_response_t.
+ */
+manager_get_device_info_response_t get_device_info(void);
```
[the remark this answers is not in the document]
ujjwal-cyph: moved

apps/manager_app/get_device_info.c
ujjwal-cyph: changed

apps/manager_app/get_device_info.c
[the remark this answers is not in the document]
ujjwal-cyph: when all the option files are defined then we can get a valid macro for the size. at the moment, `MANAGER_GET_DEVICE_INFO_RESULT_RESPONSE_SIZE` is of 510 which is sufficient for the time being. i'll add a comment to update this

amanCypherock (approved):

PR #255, opened 2023-05-18
refactor: Change function signature of event getter module

kartikCypherock:


common/core/events.h
```diff
@@ -26,30 +26,15 @@
 /*****************************************************************************
  * MACROS AND DEFINES
  *****************************************************************************/
-#define EVT_CONFIG_UI_EVT (1 << 0)
-#define EVT_CONFIG_USB_EVT (1 << 1)
-#define EVT_CONGIG_NFC_EVT (1 << 2)
+#define EVT_CONFIG_UI (1 << 0)
+#define EVT_CONFIG_USB (1 << 1)
+#define EVT_CONFIG_NFC (1 << 2)
```
amanCypherock, line 31: Just because you've already touched on this, I suggest we avoid the acronym evt and use full name event.

common/core/events.c
```diff
@@ -92,41 +92,41 @@
...
  * GLOBAL FUNCTIONS
  *****************************************************************************/
-void get_events(evt_config_t evt_config, evt_status_t *p_evt_status) {
-  if (p_evt_status == NULL) {
-    return;
-  }
+evt_status_t get_events(uint8_t event_config, uint32_t timeout) {
+  evt_status_t status = {0};
 
-  /* Configure event getters if required */
```
amanCypherock, line 100: Why remove comment?

common/core/events.h
amanCypherock: ATCG

amanCypherock (approved):

common/core/events.h
ujjwal-cyph: changed

common/core/events.c
ujjwal-cyph: seems it is not necessary to keep the comment.

PR #258, opened 2023-05-18
refactor: Add new api in engine to run single step

kartikCypherock:
This change is done to enable re-use of `engine_run_step` from different locations, where creating a buffer is overhead since there is only a single step in the flow


kartikCypherock:
Deemed as not a requirement as of now, therefore closing the PR

PR #260, opened 2023-05-19
Fix core apis: usb and application status

ujjwal-cyph:
Fixes

#### 1. Current usb_send_msg exits on empty msg
The current behavior is not consistent with the expectations of the API. Because, if the application sends 0-sized messages, the current API will do an early exit which leads to unserved USB events. This can cause issues in the application state and hence out of sync with the host application.

#### 2. Correct abort flag assignment case

### DIFF
--- a/common/core/status_api.c
+++ b/common/core/status_api.c
@@ -108,8 +108,11 @@ void core_status_set_idle_state(core_device_idle_state_t idle_state) {
   if (CORE_DEVICE_IDLE_STATE_IDLE == core_status.device_idle_state)
     comm_reset_interface();
 
-  core_status.abort_disabled =
-      (CORE_DEVICE_IDLE_STATE_USB == core_status.device_idle_state);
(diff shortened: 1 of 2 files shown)


kartikCypherock (approved):

common/core/status_api.c
```diff
@@ -108,8 +108,11 @@ void core_status_set_idle_state(core_device_idle_state_t idle_state) {
   if (CORE_DEVICE_IDLE_STATE_IDLE == core_status.device_idle_state)
     comm_reset_interface();
 
-  core_status.abort_disabled =
-      (CORE_DEVICE_IDLE_STATE_USB == core_status.device_idle_state);
+  if (CORE_DEVICE_IDLE_STATE_USB == core_status.device_idle_state) {
```
kartikCypherock, line 111:
```suggestion
  if (CORE_DEVICE_IDLE_STATE_USB != core_status.device_idle_state) {
```

kartikCypherock (approved):

PR #274, opened 2023-06-01
feat: data error handling

kartikCypherock:
Implements https://app.clickup.com/t/85zt3zq2f

### DIFF
--- a/apps/manager_app/joystick_training.c
+++ b/apps/manager_app/joystick_training.c
@@ -102,27 +102,15 @@ typedef struct joystick_step {
 /*****************************************************************************
  * STATIC FUNCTION PROTOTYPES
  *****************************************************************************/
-
 /**
- * @brief Checks validity of the entry-point training context data.
- * @details The function will ensure the following requirements:
- * <ol>
- * <li> stale p0 events do not exist at entry. </li>
- * <li> ensure data validity- received query if & is correctly decoded <li>
- * </ol>
- * In case of failure, the function will convey appropriate error to host. The
- * caller function need not handle the usb event as this function does that for
- * all the early exit cases.
+ * @brief This function returns the which_request tag of request of type
+ * manager_train_joystick_request_t
  *
- * @return Whether the check is pass or fail.
+ * @param request Reference to query
+ * @return pb_size_t The which_request tag held in the query
  */
-static bool validate_training_request(manager_query_t *query);
-
-/**
- * @brief Fill the error in the structure generated via protobuf.
- * @details The function sets the necessary fields for the data related errors.
- */
-static void send_training_error(uint32_t error_code);
+static pb_size_t get_which_request(
+    const manager_train_joystick_request_t *request);
 
 /**
  * @brief Sends the encoded message via usb api to the host.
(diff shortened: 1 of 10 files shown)


amanCypherock:
Can we break this PR such that the manager APIs are merged in 1 PR and other app functions handling are taken up 1 PR at a time?
Currently, this PR is blocked as we have to review, and test every function of the manager app and would need to do follow-ups accordingly. If we merge the APIs first, we will have an option to use these functions and merge of one flow will not block the other.

kartikCypherock:
> Can we break this PR such that the manager APIs are merged in 1 PR and other app functions handling are taken up 1 PR at a time? Currently, this PR is blocked as we have to review, and test every function of the manager app and would need to do follow-ups accordingly. If we merge the APIs first, we will have an option to use these functions and merge of one flow will not block the other.

I have reverted change from `auth_card.c`, for other files, we have already integrated new APIs, so it is not feasible to revert the change. Please re-review

apps/manager_app/manager_api.c
```diff
@@ -72,6 +75,13 @@
 /*****************************************************************************
  * PRIVATE MACROS AND DEFINES
  *****************************************************************************/
+#define P0_SESSION_HANDLER(event)                                              \
+  do {                                                                         \
+    if (true == event.p0_event.flag) {                                         \
+      manager_send_data_flow_error(ERROR_DATA_FLOW_INACTIVITY_TIMEOUT);        \
```
ujjwal-cyph, line 81: abort event should not send this error

kartikCypherock: fixed

apps/manager_app/joystick_training.c
```diff
@@ -218,7 +180,15 @@ static bool training_step(const joystick_step_t *step) {
 void manager_joystick_training(manager_query_t *query) {
   /* Validate if this flow is allowed */
   if (false == onboarding_step_allowed(ONBOARDING_JOYSTICK_TRAINING)) {
-    // TODO: Reject query
+    manager_send_data_flow_error(ERROR_DATA_FLOW_INVALID_REQUEST);
+    return;
+  }
+
+  /* Validate request */
+  if (MANAGER_TRAIN_JOYSTICK_REQUEST_INITIATE_TAG ==
```
ujjwal-cyph, line 188: this check should be inverted `!=` in place of `==`

apps/manager_app/manager_api.c
```diff
@@ -72,6 +75,15 @@
 /*****************************************************************************
  * PRIVATE MACROS AND DEFINES
  *****************************************************************************/
+#define P0_SESSION_HANDLER(event)                                              \
+  do {                                                                         \
+    if (true == event.p0_event.flag) {                                         \
+      if (true == event.p0_event.inactivity_evt) {                             \
+        manager_send_data_flow_error(ERROR_DATA_FLOW_INACTIVITY_TIMEOUT);      \
```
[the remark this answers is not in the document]
kartikCypherock: updated

ujjwal-cyph (approved):

amanCypherock:
Could you update the submodule here? The latest cypherock-common submodule commit is 95028550f962e97302aa454dab822a1bc721d17e

kartikCypherock:
> Could you update the submodule here? The latest cypherock-common submodule commit is 95028550f962e97302aa454dab822a1bc721d17e

Ok, done

amanCypherock (approved):

PR #277, opened 2023-06-01
Feat/card auth handlers/prf 4731

amanCypherock:
https://app.clickup.com/t/9002019994/PRF-4731

### DIFF
--- a/common/interfaces/card_interface/nfc_events.c
+++ b/common/interfaces/card_interface/nfc_events.c
@@ -178,6 +178,7 @@ void nfc_set_card_removed_event(void) {
 
 void nfc_en_select_card_task(void) {
   nfc_state = NFC_STATE_SET_SELECT_CARD_CMD;
+  nfc_deselect_card();
 }
 
 uint32_t nfc_en_wait_for_card_removal_task(void) {
--- a/common/interfaces/card_interface/nfc_events.c
+++ b/common/interfaces/card_interface/nfc_events.c
@@ -178,6 +178,10 @@ void nfc_set_card_removed_event(void) {
 
 void nfc_en_select_card_task(void) {
   nfc_state = NFC_STATE_SET_SELECT_CARD_CMD;
+
+  // Deselect card before selection, to avoid unexpected issues.
+  // Without deselection failure experienced on second time detection of the
+  // same card.
   nfc_deselect_card();
 }
 
--- a/src/card_operations/card_sign.c
+++ b/src/card_operations/card_sign.c
@@ -106,9 +106,8 @@ static ISO7816 get_card_auth_signature(uint8_t *sign_data,
  * @param card_data Pointer to the smart card operation data.
  * @param sign_data Pointer to the data to be signed.
  */
-static void handle_card_init_and_sign_data_operation(
-    card_operation_data_t *card_data,
-    card_sign_data_config_t *sign_data);
+static void handle_card_sign_data_operation(card_operation_data_t *card_data,
+                                            card_sign_data_config_t *sign_data);
 
 /**
  * @brief Handles the response of a smart card sign operation.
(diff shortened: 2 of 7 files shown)


rohanagarwal94:
Task linked: [PRF-4731 Card Authentication: Refactor](https://app.clickup.com/t/9002019994/PRF-4731)

src/card_operations/card_utils.c
```diff
@@ -0,0 +1,162 @@
...
+/*****************************************************************************
+ * STATIC FUNCTIONS
+ *****************************************************************************/
+
+/*****************************************************************************
+ * GLOBAL FUNCTIONS
+ *****************************************************************************/
+NFC_connection_data init_nfc_connection_data(const uint8_t *family_id,
+                                             uint8_t acceptable_cards) {
+  ASSERT(NULL != family_id);
```
ujjwal-cyph, line 105: do not assert. conditionally copy. in case of NULL, init can set 0xFF

src/card_operations/card_sign.c
```diff
@@ -0,0 +1,245 @@
...
+
+static void handle_card_init_and_sign_data_operation(
+    card_operation_data_t *card_data,
+    card_sign_data_config_t *sign_data) {
+  card_data->nfc_data = init_nfc_connection_data(sign_data->family_id,
+                                                 sign_data->acceptable_cards);
+  card_initialize_applet(card_data);
+
+  if (CARD_OPERATION_SUCCESS == card_data->error_type) {
+    if (DEFAULT_UINT32_IN_FLASH == *(uint32_t *)(sign_data->family_id)) {
```
ujjwal-cyph, line 164: `*(uint32_t *)` typecasting for the sake of reading is not good. alternatives: `U32_READ_BE_ARRAY`, `read_be`

src/card_operations/card_utils.c
```diff
@@ -0,0 +1,162 @@
...
+  card_number = decode_card_number(nfc_data->tapped_card);
+  serial[CARD_ID_SIZE - 1] = card_number;
+}
+
+card_error_type_e wait_for_card_removal(void) {
+  instruction_scr_change_text(ui_text_remove_card_prompt, true);
+  if (0 != nfc_en_wait_for_card_removal_task()) {
+    // Card not selected or removed
+    return CARD_OPERATION_SUCCESS;
+  }
```
ujjwal-cyph, line 148: check this before rendering display

amanCypherock: added

src/card_operations/card_utils.c
amanCypherock: updated

src/card_operations/card_sign.h
```diff
@@ -0,0 +1,98 @@
...
+      0                                                                        \
+    }                                                                          \
+  }
+
+/*****************************************************************************
+ * TYPEDEFS
+ *****************************************************************************/
+typedef enum card_sign_type {
+  CARD_SIGN_SERIAL = 1,
+  CARD_SIGN_CUSTOM,
```
ujjwal-cyph, line 58: this distinction is not necessary

common/interfaces/card_interface/nfc_events.c
```diff
@@ -178,6 +178,7 @@ void nfc_set_card_removed_event(void) {
 
 void nfc_en_select_card_task(void) {
   nfc_state = NFC_STATE_SET_SELECT_CARD_CMD;
+  nfc_deselect_card();
```
ujjwal-cyph, line 181: why this addition is required?

amanCypherock: If an NFC event occurs and the card is not removed, and then waiting for the NFC event, card is not detected.

Steps to recreate issue:
- Call `card_initialize_applet`
- Tap card but don't lift
- Call `card_initialize_applet` again (hence the NFC event again)

Expected behaviour is that card will not be detected on the second call.


Deselecting the card before waiting for event again solves this.

src/card_operations/card_sign.h
amanCypherock: Both cases are different in their data source.

In both cases, a data buffer is required but in the serial case, the data is set internally from family id and card number.

amanCypherock: Will add an explanation in enum declaration as well.

ujjwal-cyph: okay. this makes sense

common/interfaces/card_interface/nfc_events.c
ujjwal-cyph: is there any relevant place to add this knowledge?

amanCypherock: I'll add a comment here with the change itself.
I haven't done the complete RCA on why the selection is not working on the second event. This might even be a temporary solution.

The idea for this change is to deselect the card just before selecting it, so we don't have to deselect in every flow.

amanCypherock: added

ujjwal-cyph (approved):

PR #278, opened 2023-06-02
Added card auth sign serial and challenge handlers PRF-4731

amanCypherock:
https://app.clickup.com/t/9002019994/PRF-4731

### DIFF
--- a/apps/manager_app/auth_card.c
+++ b/apps/manager_app/auth_card.c
@@ -340,26 +340,52 @@ static manager_error_code_t prepare_card_auth_context(
   return MANAGER_TASK_SUCCESS;
 }
 
-static manager_error_code_t handle_sign_card_serial(
+static manager_error_code_t handle_sign_data(
     auth_card_data_t *auth_card_data,
     manager_auth_card_response_t *resp) {
-  instruction_scr_init(auth_card_data->ctx.message,
-                       auth_card_data->ctx.heading);
+  card_sign_data_config_t sign_config = {0};
 
-  card_sign_data_config_t sign_serial_config =
-      INIT_SIGN_SERIAL_CONFIG(auth_card_data->ctx.acceptable_cards,
-                              auth_card_data->ctx.family_id,
-                              true,
-                              auth_card_data->ctx.heading,
-                              auth_card_data->ctx.message,
-                              resp->serial_signature.serial);
-  card_error_type_e status = card_sign_auth_data(&sign_serial_config);
+  sign_config.acceptable_cards = auth_card_data->ctx.acceptable_cards;
+  sign_config.heading = auth_card_data->ctx.heading;
+  sign_config.message = auth_card_data->ctx.message;
+  sign_config.family_id = auth_card_data->ctx.family_id;
+
+  if (MANAGER_AUTH_CARD_REQUEST_INITIATE_TAG ==
+      auth_card_data->query->auth_card.which_request) {
+    sign_config.data = resp->serial_signature.serial;
+    sign_config.sign_type = CARD_SIGN_SERIAL;
+  } else if (MANAGER_AUTH_CARD_REQUEST_CHALLENGE_TAG ==
+             auth_card_data->query->auth_card.which_request) {
+    sign_config.data = auth_card_data->query->auth_card.challenge.challenge;
+    sign_config.data_size = CHALLENGE_SIZE;
+    sign_config.sign_type = CARD_SIGN_CUSTOM;
+  } else {
+    // This function shouldn't be called in other states
+    return MANAGER_TASK_INVALID_STATE;
+  }
+
+  if (MANAGER_AUTH_CARD_STATUS_SERIAL_SIGNED != core_status_get_flow_status() &&
+      false == auth_card_data->ctx.pair_card_required)
+    sign_config.skip_card_removal = false;
+  else
+    sign_config.skip_card_removal = true;
+
+  card_error_type_e status = card_sign_auth_data(&sign_config);
 
   switch (status) {
     case CARD_OPERATION_SUCCESS:
-      memcpy(resp->serial_signature.signature,
-             sign_serial_config.signature,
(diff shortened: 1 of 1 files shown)


rohanagarwal94:
Task linked: [PRF-4731 Card Authentication: Refactor](https://app.clickup.com/t/9002019994/PRF-4731)

kartikCypherock (approved):
Conditionally approving the change, based on integration testing with SDK/Cysync

apps/manager_app/auth_card.c
```diff
@@ -507,9 +549,9 @@ void card_auth_handler(manager_query_t *query) {
   ASSERT(NULL != query);
 
   /* Validate if this flow is allowed */
-  if (false == onboarding_step_allowed(ONBOARDING_CARD_AUTHENTICATION)) {
-    // TODO: Reject query
-  }
+  // if (false == onboarding_step_allowed(ONBOARDING_CARD_AUTHENTICATION)) {
+  //   // TODO: Reject query
+  // }
```
amanCypherock, line 554:
```suggestion
  if (false == onboarding_step_allowed(ONBOARDING_CARD_AUTHENTICATION)) {
    // TODO: Reject query
  }
```

apps/manager_app/auth_card.c
```diff
@@ -111,6 +112,7 @@ typedef struct {
   char heading[30];
   char message[100];
   uint8_t acceptable_cards;
+  uint8_t family_id[FAMILY_ID_SIZE];
   bool pair_card_required;
 } auth_card_screen_ctx_t;
```
kartikCypherock, line 117: Is this structure a duplicate of `card_sign_data_config_t` ?

apps/manager_app/auth_card.c
```diff
@@ -332,20 +346,33 @@ static manager_error_code_t handle_sign_card_serial(
...
+  card_error_type_e status = card_sign_auth_data(&sign_serial_config);
+
+  switch (status) {
+    case CARD_OPERATION_SUCCESS:
+      memcpy(resp->serial_signature.signature,
+             sign_serial_config.signature,
+             sizeof(resp->serial_signature.signature));
+      if (DEFAULT_UINT32_IN_FLASH == U32_READ_BE_ARRAY(get_family_id())) {
+        set_family_id_flash(auth_card_data->ctx.family_id);
+      }
```
kartikCypherock, line 365: We are already writing family ID in flash memory during "card tap check".. so are we deeming it unsafe to be there?

apps/manager_app/auth_card.c
```diff
@@ -539,6 +581,9 @@ void card_auth_handler(manager_query_t *query) {
     }
 
     send_auth_card_response(&resp);
+    if (MANAGER_AUTH_CARD_RESPONSE_COMMON_ERROR_TAG == resp.which_response) {
```
kartikCypherock, line 584: Please consider re-visiting this flow again (when time permits) to improve handling of exit conditions...

apps/manager_app/auth_card.c
amanCypherock: removed

apps/manager_app/auth_card.c
amanCypherock: `card_sign_data_config_t` takes a pointer and the `auth_card_screen_ctx_t` object member is passed to that config.

This is helpful when FF value is passed in family id.

apps/manager_app/auth_card.c
amanCypherock: Agreed, I understand that sending response and errors from here is not ideal, so we can update this as follows
- Each handler should send their own response to host for success case 
- Send error response to host in errors cases before showing error display.
- The parent should only be concerned with correct exit condition i.e. flow complete or error occurred
- Flow complete should be indicated by flow status.
- For early exits, say auth failed after the serial sign, handler can return MANAGER_TASK_FAILED to exit.

ujjwal-cyph: can add these points as TODO here

amanCypherock: will do.

kartikCypherock: Got it, please add as to do so we can move ahead

apps/manager_app/auth_card.c
```diff
@@ -324,6 +335,9 @@ static void prepare_card_auth_context(auth_card_data_t *auth_card_data) {
            sizeof(auth_card_data->ctx.message),
            UI_TEXT_PLACE_CARD_TILL_BEEP,
            SIGN_SERIAL_BEEP_COUNT(auth_card_data->ctx.pair_card_required));
+
+  memcpy(auth_card_data->ctx.family_id, get_family_id(), FAMILY_ID_SIZE);
+  return MANAGER_TASK_SUCCESS;
 }
 
 static manager_error_code_t handle_sign_card_serial(
```
ujjwal-cyph, line 343: body of `handle_sign_card_serial` and `handle_sign_challenge` is almost same. would be best if the duplication is removed.

amanCypherock: updated, please check

apps/manager_app/auth_card.c
```diff
@@ -324,28 +335,76 @@ static void prepare_card_auth_context(auth_card_data_t *auth_card_data) {
            sizeof(auth_card_data->ctx.message),
            UI_TEXT_PLACE_CARD_TILL_BEEP,
            SIGN_SERIAL_BEEP_COUNT(auth_card_data->ctx.pair_card_required));
+
+  memcpy(auth_card_data->ctx.family_id, get_family_id(), FAMILY_ID_SIZE);
+  return MANAGER_TASK_SUCCESS;
 }
 
-static manager_error_code_t handle_sign_card_serial(
+static manager_error_code_t handle_sign_data(
```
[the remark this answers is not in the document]
amanCypherock: Added

ujjwal-cyph (approved):

PR #280, opened 2023-06-03
feat(app): Updated pair operation and integrated with auth card PRF-4707 PRF-4731

amanCypherock:
https://app.clickup.com/t/9002019994/PRF-4707


rohanagarwal94:
Task linked: [CU-85zt1bm1p Card Authentication: Refactor](https://app.clickup.com/t/85zt1bm1p)

apps/manager_app/auth_card.c
```diff
@@ -496,10 +500,16 @@ static manager_error_code_t handle_auth_card_result_query(
...
-          /* TODO: Check if card operation failed*/
-          // return MANAGER_TASK_FAILED;
+          uint8_t card_number =
+              decode_card_number(auth_card_data->ctx.acceptable_cards);
+          bool result = pair_card_operation(card_number,
+                                            auth_card_data->ctx.heading,
+                                            auth_card_data->ctx.message);
+
+          if (true != result) {
+            delay_scr_init("Pairing operation failed", DELAY_TIME);
```
kartikCypherock, line 510: Can you check if having delay screen before sending response is a good idea? I think it can hamper the synchronization between cysync and device

amanCypherock: Agreed, working on that change separately.

src/level_four/tap_cards/controller/pair_card_controller.c
```diff
@@ -140,48 +152,82 @@ static void _tap_card_backend(uint8_t card_number) {
...
-      instruction_scr_change_text(ui_text_card_removed_fast, true);
-    if (!tap_card_applet_connection())
-      return;
-    if (is_paired(tap_card_data.card_key_id) > -1) {
-      instruction_scr_destructor();
-      flow_level.level_four++;
-      return;
+    // Initialize card tap config
+    card_data.nfc_data.acceptable_cards = (1 << (card_number - 1));
+    card_data.nfc_data.lvl3_retry_point = flow_level.level_three - 1;
```
kartikCypherock, line 172:
```suggestion
```

src/level_four/tap_cards/controller/pair_card_controller.c
```diff
@@ -121,6 +124,9 @@ void tap_card_pair_card_controller() {
 }
 
 static void _tap_card_backend(uint8_t card_number) {
+}
```
kartikCypherock, line 127: Add a comment to tell which function to use instead of `_tap_card_backend`

src/level_four/tap_cards/controller/pair_card_controller.c
```diff
@@ -140,48 +152,82 @@ static void _tap_card_backend(uint8_t card_number) {
...
-        set_family_id_flash(tap_card_data.family_id);
-      instruction_scr_change_text(ui_text_remove_card_prompt, true);
-      handle_pair_card_success(card_number, session_nonce, card_pairing_data);
-      nfc_detect_card_removal();
-      instruction_scr_destructor();
+      if (CARD_OPERATION_SUCCESS != wait_for_card_removal()) {
+        result = false;
+      } else {
+        result = true;
+      }
```
kartikCypherock, line 217: Is card removal going to be an unrecoverable error?

src/level_four/tap_cards/controller/pair_card_controller.c
```diff
@@ -140,48 +152,82 @@ static void _tap_card_backend(uint8_t card_number) {
...
+      buzzer_start(BUZZER_DURATION);
+      if (CARD_OPERATION_SUCCESS !=
+          display_error_message(card_data.error_message)) {
+        result = false;
+        break;
+      }
+
+      // Reset instruction screen if retap required
+      if (CARD_OPERATION_RETAP_BY_USER_REQUIRED == card_data.error_type) {
+        instruction_scr_init(message, heading);
```
kartikCypherock, line 207: Let this be the first statement inside while(1) loop so that we don't need to worry about this.. in case any error occurs and we have to retry we will not have to worry..

Saying because we missed calling this in line 221,222,223 where retap was required

src/level_four/tap_cards/controller/pair_card_controller.c
amanCypherock: only checking P0 which will be irrecoverable

src/level_four/tap_cards/controller/pair_card_controller.c
amanCypherock: We want to set the instruction screen on start of the flow and in while(1) when recoverable error is set.

For this requirement, we can add the instruction screen just before entry to while loop.

src/level_four/tap_cards/controller/pair_card_controller.c
kartikCypherock: No I do not thing considering P0 event in card removal is required.

src/level_four/tap_cards/controller/pair_card_controller.c
amanCypherock: removed

src/level_four/tap_cards/controller/pair_card_controller.c
amanCypherock: updated

src/level_four/tap_cards/controller/pair_card_controller.c
amanCypherock: Updated the hanlding.

apps/manager_app/auth_card.c
kartikCypherock: Ok, resolving for now

kartikCypherock (approved):

PR #282, opened 2023-06-06
feat (core): Enable support for log export to host

ujjwal-cyph:


common/logger/logger.h
```diff
@@ -102,7 +102,7 @@ void logger(char *fmt, ...);
  *
  * @note
  */
-void logger_task(void);
+void logger_task(uint8_t *data, size_t *size);
```
kartikCypherock, line 105: Please complete doxygen comments

ujjwal-cyph: added

kartikCypherock (approved):

PR #297, opened 2023-06-15
fix: Card training buzzer and state correction

ujjwal-cyph:
The buzzer should make a sound in the success case only. Abort from the host should not trigger the buzzer. Also, card operation abort errors are now handled globally hence its handling is being removed. Update the card training state to the flash.

### DIFF
--- a/src/card_operations/check_pairing.c
+++ b/src/card_operations/check_pairing.c
@@ -126,10 +126,10 @@ card_error_type_e card_check_pairing(check_pairing_result_t *result) {
   card_operation_data_t operation_data = {{0}, NULL, 0};
   init_tap_card_data(&operation_data.nfc_data);
   status = card_initialize_applet(&operation_data);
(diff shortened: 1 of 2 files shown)


src/card_operations/check_pairing.c
```diff
@@ -126,10 +126,10 @@ card_error_type_e card_check_pairing(check_pairing_result_t *result) {
   card_operation_data_t operation_data = {{0}, NULL, 0};
   init_tap_card_data(&operation_data.nfc_data);
   status = card_initialize_applet(&operation_data);
-  buzzer_start(100);
   nfc_deselect_card();
 
   if (CARD_OPERATION_SUCCESS == status) {
+    buzzer_start(100);
```
kartikCypherock, line 132:
```suggestion
    buzzer_start(BUZZER_DURATION);
```

kartikCypherock (approved):

PR #304, opened 2023-06-16
feat: Add manager function for device firmware update confirmation

kartikCypherock:
Implements https://app.clickup.com/t/9002019994/PRF-4822

### DIFF
--- a/apps/manager_app/firmware_update.c
+++ b/apps/manager_app/firmware_update.c
@@ -98,6 +98,16 @@
 static bool check_which_request(const manager_query_t *query,
                                 pb_size_t which_request);
 
+/**
+ * @brief Checks if the provided request has valid data
+ * @details It checks if the target firmware version is strictly greater than
+ * the currently installed firmware version
+ *
+ * @param request Reference to the request received from the host
+ * @return true Indicating that the request is valid
+ * @return false Indicating that the request is not valid
+ */
+static bool validate_query(manager_firmware_update_request_t *request);
 /*****************************************************************************
  * STATIC VARIABLES
  *****************************************************************************/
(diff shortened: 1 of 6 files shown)


rohanagarwal94:
Task linked: [PRF-4822 Manager App - Device update](https://app.clickup.com/t/9002019994/PRF-4822)

apps/manager_app/firmware_update.c
```diff
@@ -0,0 +1,167 @@
...
+          query, MANAGER_FIRMWARE_UPDATE_INITIATE_REQUEST_VERSION_TAG)) {
+    return;
+  }
+
+  if (false == query->firmware_update.initiate.has_version) {
+    manager_send_error(ERROR_COMMON_ERROR_CORRUPT_DATA_TAG,
+                       ERROR_DATA_FLOW_FIELD_MISSING);
+    return;
+  }
+
```
amanCypherock, line 137: We should check if the firmware version is lower or not before starting a firmware update here.
Also, reduce the size of the major minor and patch version in the options file.

common/Firewall/sec_flash.h
```diff
@@ -18,6 +18,9 @@
 #ifndef SECURETASKS_PROTECTED_CODE_H_
 #define SECURETASKS_PROTECTED_CODE_H_
 
+#include "flash_config.h"
+#include "wallet.h"
+
```
amanCypherock, line 23: Since we haven't made any changes to this file, can you mention the intent of adding these includes?

apps/manager_app/firmware_update.c
kartikCypherock: 1. Added options file for firmware version structure
2. Added check to prevent firmware version rollback

common/Firewall/sec_flash.h
kartikCypherock: These includes are required in this header file, otherwise including sec_flash.h by itself results in build failures.

Refer : https://github.com/Cypherock/x1_wallet_firmware/actions/runs/5308445320/jobs/9607939780

apps/manager_app/firmware_update.c
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * @brief Checks if the provided request has valid data
+ * @details It checks if the target firmware version is strictly greater than
+ * the currently installed firmware version
+ *
+ * @param request Reference to the request received from the host
+ * @return true Indicating that the request is valid
+ * @return false Indicating that the request is not valid
+ */
+static bool validate_query(manager_firmware_update_request_t *request);
```
ujjwal-cyph, line 110: can keep the parameter as `const`

apps/manager_app/firmware_update.c
```diff
@@ -0,0 +1,195 @@
...
+  resp->which_response = MANAGER_FIRMWARE_UPDATE_RESPONSE_CONFIRMED_TAG;
+  resp->confirmed.dummy_field = '\0';
+  manager_send_result(&result);
+
+  // NOTE: This is a USB initiated flow, however, device will go in bootloader
+  // mode after blocking delay of 500ms without serving any events. So in case
+  // any abort event is triggered by the host, it will NOT be served!
+  // Wait for status pull to desktop (which requests at 200ms)
+  instruction_scr_init(ui_text_processing, NULL);
+  BSP_DelayMs(500);
```
ujjwal-cyph, line 190: can add following comment.

`// hard delay to save from glitchy behaviour`

apps/manager_app/firmware_update.c
kartikCypherock: Updated

apps/manager_app/firmware_update.c
kartikCypherock: Hey, while you are right, but this delay is not to prevent glitchy behaviour, we could have used delay screen for that.

Delay is to ensure that user confirmation response is polled by the host which is done periodically every 200ms.

amanCypherock (approved):

PR #328, opened 2023-07-06
fix (ui): Clear object references across `ui_instruction_scr` instances

ujjwal-cyph:
The `lv_object` references stored in the static objects are not cleared before the next instantiation of the same UI. This can cause some logic to fail as specific decisions are made on the existence of those references.

### DIFF
--- a/common/interfaces/user_interface/ui_instruction.c
+++ b/common/interfaces/user_interface/ui_instruction.c
@@ -92,7 +92,7 @@ void instruction_scr_init(const char *message, const char *heading_text) {
   }
   instruction = lv_label_create(lv_scr_act(), NULL);
   ui_paragraph(instruction, message, LV_LABEL_ALIGN_CENTER);
-  if (heading_text == NULL)
+  if (NULL == heading_text)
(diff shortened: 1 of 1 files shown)


kartikCypherock (approved):

common/interfaces/user_interface/ui_instruction.c
```diff
@@ -87,11 +91,8 @@ void instruction_scr_init(const char *message, const char *heading_text) {
                LV_LABEL_ALIGN_CENTER);
   }
   instruction = lv_label_create(lv_scr_act(), NULL);
-  ui_paragraph(
-      instruction,
-      message,
-      LV_LABEL_ALIGN_CENTER);    // Creates task to print text on screen
-  if (heading == NULL)
+  ui_paragraph(instruction, message, LV_LABEL_ALIGN_CENTER);
+  if (heading_text == NULL)
```
amanCypherock, line 95:
```suggestion
  if (NULL == heading_text)
```

amanCypherock (approved):

PR #335, opened 2023-07-11
refactor: Port card operation to view card version and enable setting

kartikCypherock:
This pull request enables viewing X1 card version. A new field `card_version` is added in `nfc_connection_data` structure, which can hold a pointer to buffer where the card version will be copied on card tap.

Fixes https://app.clickup.com/t/9002019994/PRF-4845

### DIFF
--- a/src/card_operations/read_card_version.c
+++ b/src/card_operations/read_card_version.c
@@ -103,6 +103,7 @@ bool read_card_version(uint8_t *card_version,
   card_operation_data_t card_data = {0};
   card_data.nfc_data.retries = 5;
   card_data.nfc_data.card_version = card_version;
+  card_data.nfc_data.init_session_keys = false;
 
   instruction_scr_init(msg, heading);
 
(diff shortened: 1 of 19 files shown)


rohanagarwal94:
Task linked: [PRF-4845 Card version](https://app.clickup.com/t/9002019994/PRF-4845)

src/card_operations/read_card_version.c
```diff
@@ -0,0 +1,140 @@
...
+                       const char *heading,
+                       const char *msg) {
+  bool result = false;
+  card_operation_data_t card_data = {0};
+  card_data.nfc_data.retries = 5;
+  card_data.nfc_data.card_version = card_version;
+
+  instruction_scr_init(msg, heading);
+
+  while (1) {
```
ujjwal-cyph, line 109: can we remove the `while...` as the retriable situations are already handled by `card_initialize_applet`

kartikCypherock: Hi Ujjwal, this while loop is required to handle only `CARD_OPERATION_CARD_REMOVED` error type, which would require a retap from the user..

ujjwal-cyph: okay. makes sense

ujjwal-cyph (approved):

PR #338, opened 2023-07-15
feat(core): Added old wallet menu

amanCypherock:
Adds old wallet menu navigation only.

PR #339, opened 2023-07-15
Refactor/view seed flow/prf 4839

amanCypherock:


rohanagarwal94:
Task linked: [PRF-4839 Task controller merge - View wallet seed](https://app.clickup.com/t/9002019994/PRF-4839)

amanCypherock:
This PR is dependent on https://github.com/Cypherock/x1_wallet_firmware/pull/338

src/wallet/reconstruct_seed_flow.c
```diff
@@ -281,25 +279,68 @@ bool reconstruct_seed_flow(const uint8_t *wallet_id, uint8_t *seed_out) {
...
+  uint8_t result = 0;
+
+  if (COMPLETED == handle_secret_reconstruction(wallet_id, secret, PIN_INPUT)) {
+    mnemonic_clear();
+    const char *mnemo =
+        mnemonic_from_data(secret, wallet.number_of_mnemonics * 4 / 3);
+    uint16_t single_line_mnemonics_length =
+        strnlen(mnemo, MAX_NUMBER_OF_MNEMONIC_WORDS * MAX_MNEMONIC_WORD_LENGTH);
+    __single_to_multi_line(mnemo, single_line_mnemonics_length, mnemonic_list);
+    result = wallet.number_of_mnemonics;
```
kartikCypherock, line 337: We are potentially making a double copy of mnemonics.. can you re-check if this is required? I think this copy of mnemonics is not getting cleared as well.. 

We can move this logic in `view_seed_flow` and prevent double copy and therefore risk of missing out anything

src/wallet/reconstruct_seed_flow.c
```diff
@@ -129,8 +129,18 @@ reconstruct_state_e reconstruct_seed_handler(reconstruct_state_e state,
 /*****************************************************************************
  * STATIC FUNCTIONS
  *****************************************************************************/
+void get_seed_from_secret(uint8_t *secret, uint8_t *seed_out) {
```
kartikCypherock, line 132: Rather than breaking this into a completely different function.. can we split the state of the `reconstruct_seed_handler` - one in which mnemonics are generated and other in which seed is calculated using passphrase.. 

So something similar to `init_state` in `handle_secret_reconstruction` can be applied to `exit_state`

src/wallet/reconstruct_seed_flow.c
amanCypherock: It is not a double copy as the of mnemonics as `mnemo` is a pointer for a static variable with single line mnemonics, and we need to convert the mnemonics to multi-line, we wouldn't want to change the pointer returned by `mnemonic_from_data`.

amanCypherock: To the second comment we can move this to view_seed_flow too, but it would still need a function that returns the secret for mnemonics construction, I think it's better to keep that logic abstracted as is.

src/wallet/reconstruct_wallet_flow.c
```diff
@@ -261,14 +292,10 @@ reconstruct_state_e reconstruct_seed_handler(reconstruct_state_e state,
...
 
-/*****************************************************************************
- * GLOBAL FUNCTIONS
- *****************************************************************************/
-bool reconstruct_seed_flow(const uint8_t *wallet_id, uint8_t *seed_out) {
-  if ((NULL == wallet_id) || (NULL == seed_out)) {
-    return false;
-  }
-
+reconstruct_state_e reconstruct_wallet_secret_flow(
```
kartikCypherock, line 295: static keyword is somehow missing in static functions present in these files. Can you please add

src/wallet/reconstruct_seed_flow.c
kartikCypherock: In that case, `PASSPHRASE_INPUT` state should also not be in same handler. It is introducing an inconsistency, where we have flexibility to have configurable init state but not configuration exit state

src/wallet/reconstruct_wallet_flow.c
```diff
@@ -261,14 +292,10 @@ reconstruct_state_e reconstruct_seed_handler(reconstruct_state_e state,
...
+    const uint8_t *wallet_id,
+    uint8_t *secret_out,
+    reconstruct_state_e init_state) {
   // TODO: Consolidate in one function
   // Clear confidential data irrespective of the result of the flow
   memzero(&wallet, sizeof(wallet));
```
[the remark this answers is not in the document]
amanCypherock: I agree, but didn't do this coz it's still isn't in a single function.

Will update with a single function for clearing this data.

src/wallet/view_seed_flow.c
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * @return the next state of the view_seed_states_e enum type.
+ */
+view_seed_states_e view_seed_handler(view_seed_states_e current_state,
+                                     const uint8_t *wallet_id);
+
+/*****************************************************************************
+ * STATIC VARIABLES
+ *****************************************************************************/
+CONFIDENTIAL char mnemonics[MAX_NUMBER_OF_MNEMONIC_WORDS]
```
ujjwal-cyph, line 113: if we can remove the states/switch cases, this global variable can be dissolved.

src/wallet/view_seed_flow.c
```diff
@@ -0,0 +1,205 @@
...
+          ui_text_view_seed_messages, 3, DELAY_LONG_STRING, true);
+      next_state = get_state_on_confirm_scr(
+          VIEW_SEED_MNEMONIC_RECONSTRUCTION, EXIT_FAILURE, VIEW_SEED_TIMED_OUT);
+      break;
+
+    case VIEW_SEED_MNEMONIC_RECONSTRUCTION:
+      if (NULL == wallet_id) {
+        next_state = VIEW_SEED_COMPLETED_WITH_ERRORS;
+        break;
+      }
```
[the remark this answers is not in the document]
amanCypherock: removed and refactored

src/wallet/view_seed_flow.c
amanCypherock: refactored

src/menu/wallet_menu.c
```diff
@@ -126,15 +127,15 @@ static void ignore_p0_handler(engine_ctx_t *ctx,
...
+static uint8_t wallet_id[WALLET_ID_SIZE] = {0};
 
 static const flow_step_t wallet_step = {.step_init_cb = wallet_menu_initialize,
                                         .p0_cb = ignore_p0_handler,
                                         .ui_cb = wallet_menu_handler,
                                         .usb_cb = NULL,
                                         .nfc_cb = NULL,
                                         .evt_cfg_ptr = &device_nav_evt_config,
-                                        .flow_data_ptr = &wallet_index};
+                                        .flow_data_ptr = &wallet_id};
```
kartikCypherock, line 138: Is this required?

src/wallet/view_seed_flow.c
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+/*****************************************************************************
+ * INCLUDES
+ *****************************************************************************/
+#include "constant_texts.h"
+#include "core_error.h"
+#include "reconstruct_wallet_flow.h"
+#include "ui_delay.h"
+#include "ui_list.h"
```
kartikCypherock, line 66:
```suggestion
#include "ui_screens.h"
```

src/wallet/view_seed_flow.c
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/*****************************************************************************
+ * GLOBAL FUNCTIONS
+ *****************************************************************************/
+void view_seed_flow(uint8_t *wallet_id) {
+  ASSERT(NULL != wallet_id);
+
+  if (false == view_seed_handler(wallet_id)) {
+    mark_core_error_screen(ui_text_something_went_wrong);
```
kartikCypherock, line 146: Instead of adding error here.. we should add this error screen in reconstruct_wallet_flow handler.. in case it reaches state `COMPLETED_WITH_ERRORS`.. so that this handling is done in every flow which calls the reconstruct_wallet_flow

src/wallet/view_seed_flow.c
```diff
@@ -0,0 +1,150 @@
...
+                             [MAX_MNEMONIC_WORD_LENGTH] = {0};
+  CONFIDENTIAL uint8_t no_of_mnemonics = 0;
+  bool result = false;
+
+  do {
+    multi_instruction_init(
+        ui_text_view_seed_messages, 3, DELAY_LONG_STRING, true);
+    if (0 != get_state_on_confirm_scr(0, 1, 2)) {
+      break;
+    }
```
kartikCypherock, line 116: Reconsider replacing this with confirmation screen instead..

src/wallet/reconstruct_wallet_flow.c
amanCypherock: updated, seems like I missed to comment here.

src/menu/wallet_menu.c
amanCypherock: I was not using the flow_data_ptr earlier, added usage to initialize and handler functions. 
I think it's better to keep as many APIs as possible independent of global variables.

src/wallet/view_seed_flow.c
amanCypherock: I'll add a todo, and we can take it up as a follow up change.

src/constant_texts.c
```diff
@@ -377,10 +377,9 @@ const char *ui_text_creation_failed_delete_wallet =
     "Wallet creation failed\nProceed to delete";
 
 const char *ui_text_successfull = "Successful";
-const char *ui_text_view_seed_messages[] = {
-    "Do not give the seed phrase to anyone",
-    "You will lose your coins if the seed phrase gets stolen",
-    "Press the center button to proceed"};
+const char *ui_text_view_seed_messages =
+    "Do not give the seed phrase to anyone. You will lose your coins if the "
+    "seed phrase gets stolen";
```
kartikCypherock, line 382: Add: "Do you want to proceed?" in the message

src/wallet/view_seed_flow.c
```diff
@@ -102,15 +99,14 @@ static bool view_seed_handler(const uint8_t *wallet_id);
...
-  bool result = false;
 
   do {
-    multi_instruction_init(
-        ui_text_view_seed_messages, 3, DELAY_LONG_STRING, true);
+    ui_scrollable_page(
+        NULL, ui_text_view_seed_messages, MENU_SCROLL_HORIZONTAL, true);
     if (0 != get_state_on_confirm_scr(0, 1, 2)) {
       break;
     }
```
kartikCypherock: Replace with `core_scroll_page`

src/wallet/view_seed_flow.c
amanCypherock: updated

src/wallet/view_seed_flow.c
amanCypherock: updated

src/wallet/view_seed_flow.c
amanCypherock: updated

src/constant_texts.c
amanCypherock: added

src/wallet/view_seed_flow.c
amanCypherock: updated

kartikCypherock (approved):

ujjwal-cyph (approved):

PR #356, opened 2023-07-30
fix(nfc): Sound buzzer on nfc only

amanCypherock:


src/card_operations/card_fetch_share.c
```diff
@@ -164,8 +164,13 @@ card_error_type_e card_fetch_share(card_fetch_share_cfg_t *config) {
       }
     }
 
-    if ((CARD_OPERATION_CARD_REMOVED == card_data.error_type) ||
-        (CARD_OPERATION_RETAP_BY_USER_REQUIRED == card_data.error_type)) {
+    if (CARD_OPERATION_CARD_REMOVED == card_data.error_type) {
+      continue;
+    } else {
+      buzzer_start(BUZZER_DURATION);
```
kartikCypherock, line 170: Can you review if the buzzer can be moved outside the while 1 loop altogether?

In that case, we can remove it from the success case as well.. and have it in one common place

src/card_operations/card_delete_share.c
```diff
@@ -163,8 +164,7 @@ card_error_type_e card_delete_share(card_delete_share_cfg_t *delete_config) {
       }
     }
 
-    if (CARD_OPERATION_CARD_REMOVED == card_data.error_type ||
-        CARD_OPERATION_RETAP_BY_USER_REQUIRED == card_data.error_type) {
+    if (CARD_OPERATION_RETAP_BY_USER_REQUIRED == card_data.error_type) {
```
amanCypherock, line 167: Doesn't this remove the condition for continuing when a `CARD_OPERATION_CARD_REMOVED` error occurs?

kartikCypherock: 'CARD_OPERATION_CARD_REMOVED' will never be returned from card internal

amanCypherock: It will be returned from `card_handle_errors` it needs to be handled in the caller's while loop. This is not required for flows which only require applet initialization but for other cases, teardown can occur in middle of data exchange.

kartikCypherock: Reverted, please re-review

ujjwal-cyph (approved):

PR #358, opened 2023-08-04
refactor(flow): Port wallet edge cases to wallet menu flow

amanCypherock:


rohanagarwal94:
Task linked: [PRF-4970 Wallet edge cases redirection](https://app.clickup.com/t/9002019994/PRF-4970)

src/menu/main_menu.c
```diff
@@ -184,18 +187,15 @@ static void ignore_p0_handler(engine_ctx_t *ctx,
   ignore_p0_event();
 }
 
-static void get_wallet_id_from_list_selection(uint8_t selection,
-                                              uint8_t *wallet_id) {
-  ASSERT(0 != selection && 4 >= selection);
+static void get_flash_wallet_from_list_selection(uint8_t selection,
+                                                 Flash_Wallet **flash_wallet) {
+  ASSERT(0 != selection && 4 >= selection && NULL != flash_wallet);
```
kartikCypherock, line 192: Check the range for selection. I think it can be 4 (1 indexed)

src/menu/wallet_menu.c
```diff
@@ -120,23 +122,24 @@ static void wallet_menu_handler(engine_ctx_t *ctx,
...
+ * object of selected wallet
  */
 static void ignore_p0_handler(engine_ctx_t *ctx,
                               p0_evt_t p0_evt,
                               const void *data_ptr);
 /*****************************************************************************
  * STATIC VARIABLES
  *****************************************************************************/
-static uint8_t wallet_id[WALLET_ID_SIZE] = {0};
+static Flash_Wallet flash_wallet = {0};
```
kartikCypherock, line 134: Can you check if there is already a global variable of type `Flash_Wallet` present in the code? Aim should be to reduce duplicity..

Also, do you think we require the complete structure? Like pow_challenge` data? Or just the states/ wallet id/ name

amanCypherock: Picked the `Flash_Wallet` structure as it contains all important state variables. The function `get_flash_wallet_from_list_selection` returns the pointer to global variable of that Flash_Wallet object in flash_ram_instance.

We don't require the complete structure, the share and challenge are unnecessary for this purpose.

src/menu/main_menu.c
amanCypherock: No change has been made to the range of selection. The assert checks for 1 to 4 inclusive.

src/menu/wallet_menu.c
kartikCypherock: In that case saving a copy of Flash_Wallet seems duplicate.. as we can fetch from existing functions everytime we require the info.. not making this comment blocking though..

kartikCypherock (approved):

src/menu/wallet_menu.c
amanCypherock: In that case, would it work to just keep the flash_wallet object pointer for that wallet?

amanCypherock: updated

kartikCypherock:
LGTM

PR #368, opened 2023-08-11
refactor (app): Indicate immutability of params with const

ujjwal-cyph:
Changes
- Indicate immutability of params with const
- Remove local instance from encode api

### DIFF
--- a/apps/btc_family/btc_api.c
+++ b/apps/btc_family/btc_api.c
@@ -128,7 +128,7 @@ bool decode_btc_query(const uint8_t *data,
   return status;
 }
 
-bool encode_btc_result(btc_result_t *result,
+bool encode_btc_result(const btc_result_t *result,
                        uint8_t *buffer,
                        uint16_t max_buffer_len,
                        size_t *bytes_written_out) {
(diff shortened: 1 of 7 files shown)


apps/btc_family/btc_api.c
```diff
@@ -108,27 +108,22 @@ bool decode_btc_query(const uint8_t *data,
...
 
-  /* Initialize bitcoin query */
-  btc_query_t query = BTC_QUERY_INIT_ZERO;
-
   /* Create a stream that reads from the buffer. */
   pb_istream_t stream = pb_istream_from_buffer(data, data_size);
 
   /* Now we are ready to decode the message. */
-  bool status = pb_decode(&stream, BTC_QUERY_FIELDS, &query);
+  bool status = pb_decode(&stream, BTC_QUERY_FIELDS, query_out);
```
kartikCypherock, line 115: Let's zeroise `query_out` buffer before decoding to be safe

ujjwal-cyph: added

kartikCypherock (approved):

PR #369, opened 2023-08-11
fix: Add 'Processing' screen during heavy processing

ujjwal-cyph:


apps/btc_family/btc_txn.c
```diff
@@ -322,6 +322,9 @@ static bool handle_initiate_query(const btc_query_t *query) {
   if (!core_confirmation(msg, btc_send_error)) {
     return false;
   }
+  // show processing screen for a minimum duration (additional time added due to
+  // actual processing)
+  delay_scr_init(ui_text_processing, DELAY_SHORT);
```
kartikCypherock, line 327: Check if it's suitable to add delay after setting flow status

ujjwal-cyph: would be best. moved after sending response.

kartikCypherock (approved):

PR #381, opened 2023-08-19
fix: Do not show success screen in case of early exit

kartikCypherock:


apps/btc_family/btc_xpub.c
```diff
@@ -302,10 +303,15 @@ void btc_get_xpub(btc_query_t *query) {
...
   } else {
     // send unknown error; do not know failure reason
     btc_send_error(ERROR_COMMON_ERROR_UNKNOWN_ERROR_TAG, 1);
   }
+
+  if (!status) {
+    return;
+  }
+
   delay_scr_init(ui_text_check_cysync_app, DELAY_TIME);
```
natto1784, line 316: Nice catch! Just a nitpick, this can be changed to this to minimize obfuscation
```suggestion
  if (status) {
    delay_scr_init(ui_text_check_cysync_app, DELAY_TIME);
  }
```

natto1784 (approved):
lgtm

apps/btc_family/btc_xpub.c
kartikCypherock: thanks, I have updated

PR #388, opened 2023-08-23
feat: Enable startup checks for provisioning, wallet state and faults

kartikCypherock:
Implements https://app.clickup.com/t/9002019994/PRF-4880

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -159,13 +159,35 @@ int main(void) {
     engine_run(main_engine_ctx);
   }
 #else /* RUN_ENGINE */
-
   while (true) {
     if (keypad_get_key() != 0)
       reset_inactivity_timer();
     proof_of_work_task();
-  }
+
+#if USE_SIMULATOR == 1
+    usbsim_continue_loop();
+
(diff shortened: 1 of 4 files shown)


rohanagarwal94:
Task linked: [PRF-4880 Activate device startup checks](https://app.clickup.com/t/9002019994/PRF-4880)

src/main.c
```diff
@@ -147,99 +136,36 @@ int main(void) {
...
-#elif X1WALLET_INITIAL
-      level_one_tasks_initial();
-#else
-#error Specify what to build (X1WALLET_INITIAL or X1WALLET_MAIN)
-#endif
-    }
-
-#if USE_SIMULATOR == 1
-    usbsim_continue_loop();
-#endif    // USE_SIMULATOR
```
amanCypherock, line 211: Since we don't use the usb rn with simulator, we can remove this, but let's restore this just for reference for when we have to move it to the get events task.

kartikCypherock: Restored

amanCypherock (approved):

PR #393, opened 2023-08-25
fix: Refresh timeout counter in case joystick movement is detected

kartikCypherock:
Implements https://app.clickup.com/t/9002019994/PRF-5022


rohanagarwal94:
Task linked: [PRF-5022 Fix timeout calculation - from last joystick operation](https://app.clickup.com/t/9002019994/PRF-5022)

PR #394, opened 2023-08-25
feat: Implement restricted app, host interface and enable startup check

kartikCypherock:


apps/manager_app/manager_app.c
```diff
@@ -154,3 +153,43 @@ void manager_app_main(usb_event_t usb_evt) {
...
+      device_authentication_flow(&query);
+      break;
+    }
+    case MANAGER_QUERY_FIRMWARE_UPDATE_TAG: {
+      manager_confirm_firmware_update(&query);
+      break;
+    }
+    case MANAGER_QUERY_GET_WALLETS_TAG:
+    case MANAGER_QUERY_AUTH_CARD_TAG:
+    case MANAGER_QUERY_GET_LOGS_TAG:
```
ujjwal-cyph, line 183: i think this (`MANAGER_QUERY_GET_LOGS_TAG`) is also enabled. so let's keep same ig.

ujjwal-cyph (changes requested):

apps/manager_app/manager_app.c
kartikCypherock: good catch, fixed

ujjwal-cyph (approved):
LGTM

src/restricted_app/restricted_host_interface.c
```diff
@@ -0,0 +1,131 @@
...
+      break;
+    }
+  }
+
+  /* Device authentication is complete, reset the flow as the core will now need
+   * to render the main menu */
+  if (DEVICE_AUTHENTICATED == get_auth_state()) {
+    // this is an ideally good place to show congratulation message upon
+    // onboarding completion
+    delay_scr_init(ui_text_check_cysync_app, DELAY_TIME);
```
amanCypherock, line 126: Remove stale comment

amanCypherock (approved):

src/restricted_app/restricted_host_interface.c
kartikCypherock: I have removed the comment

PR #398, opened 2023-08-29
fix: Update user facing error message to indicate wallet locked

kartikCypherock:
Implements https://app.clickup.com/t/9002019994/PRF-6168


rohanagarwal94:
Task linked: [PRF-6168 On device add "Wrong Pin!" instead of just "Wrong!"](https://app.clickup.com/t/9002019994/PRF-6168)

PR #410, opened 2023-09-22
fix(ui): Timeout not updated on joystick activity

amanCypherock:
Fixed timeout count not being reset on joystick movement


ujjwal-cyph (approved):
LGTM

PR #445, opened 2023-11-27
feat(core): Verify fetched wallets

amanCypherock:
Part 2 of wallet verification through nonce. With this change, every time a wallet is fetched, its nonce would be verified with either a wallet from another card or a wallet already saved on the device.

Changes
- On sync flow, 2 wallets fetched would be verified for name, ID, info, and nonce between before storing the 5th share on the device.
- On any reconstruction flow, the fetched wallet would be compared with the wallet present on flash for name, id, info, and nonce.

### DIFF
--- a/src/card_operations/card_fetch_share.c
+++ b/src/card_operations/card_fetch_share.c
@@ -224,7 +224,8 @@ card_error_type_e card_fetch_share(const card_fetch_share_config_t *config,
         remaining_cards = card_data.nfc_data.acceptable_cards;
         if (!_handle_retrieve_wallet_success(config->xcor)) {
           card_data.error_type = CARD_OPERATION_VERIFICATION_FAILED;
-          mark_core_error_screen("Wallet verification failed", true);
+          mark_core_error_screen(
+              ui_text_wallet_verification_failed_in_reconstruction, true);
           break;
         }
 
--- a/src/card_operations/card_fetch_share.c
+++ b/src/card_operations/card_fetch_share.c
@@ -161,6 +161,7 @@ static bool verify_fetched_wallet(uint8_t xcor) {
 
 static bool _handle_retrieve_wallet_success(uint8_t xcor) {
   if (!verify_fetched_wallet(xcor)) {
+    LOG_ERROR("Verification failed xxx39");
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


rohanagarwal94:
Task linked: [PRF-6557 Verify wallet nonce on fetching a wallet](https://app.clickup.com/t/9002019994/PRF-6557)

rohanagarwal94:
Task linked: [PRF-6203 Update create wallet and delete wallet flows to resolve apdu len error](https://app.clickup.com/t/9002019994/PRF-6203)

src/card_operations/card_fetch_share.c
```diff
@@ -105,16 +116,66 @@ static uint8_t remaining_cards;
...
-  if (WALLET_IS_ARBITRARY_DATA(wallet.wallet_info))
+
+static bool verify_fetched_wallet(uint8_t xcor) {
+  Flash_Wallet *flash_wallet = NULL;
+  bool status =
+      get_flash_wallet_by_name((const char *)wallet.wallet_name, &flash_wallet);
+
+  ASSERT(SUCCESS == status && NULL != flash_wallet);
+
+  bool compare_status = compare_status =
```
ujjwal-cyph, line 127: something is off here.

src/card_operations/card_fetch_share.c
```diff
@@ -105,16 +116,66 @@ static uint8_t remaining_cards;
...
+    // wallet nonce.
+    uint8_t wallet_nonce[NONCE_SIZE] = {0};
+    ASSERT(SUCCESS ==
+           get_flash_wallet_nonce_by_name(
+               (const char *)flash_wallet->wallet_name, wallet_nonce));
+    compare_status &=
+        (0 == memcmp(wallet.wallet_share_with_mac_and_nonce + BLOCK_SIZE,
+                     wallet_nonce,
+                     NONCE_SIZE));
+  } else if (0 < xcor) {
```
ujjwal-cyph, line 151: generalize the `else-if` to `else` only.

Add assertion/separate-check on xcor instead.

ujjwal-cyph (changes requested):
change according to the comments.

src/card_operations/card_fetch_share.c
amanCypherock: Consider the confusion when renaming as card number.
This is not technically the card number we're comparing here, of the 2 cards tapped on device, this variable represents the 2nd card tapped regardless of the #cardNumber.

src/card_operations/card_fetch_share.c
amanCypherock: updated

src/card_operations/card_fetch_share.c
amanCypherock: updated the doxygen comments here for the variable.

ujjwal-cyph (approved):
LGTM

PR #465, opened 2024-01-22
fix (core): Set default pow-nonce to 0xff in sync wallets

ujjwal-cyph:
The nonce value indicates if the wallet unlocking is initialized. The value 0xff...ff (32-bytes) is an indicative of uninitialized state while any other value will indicate progress of the unlocking process. Currently, The sync-wallet flow sets it to 0 which causes the unlock flow to skip card tapping step from the user.
Fixes https://app.clickup.com/t/9002019994/PRF-6961


rohanagarwal94:
Task linked: [PRF-6961 On device locked wallet is not getting unlocked after restore from cards](https://app.clickup.com/t/9002019994/PRF-6961)

cypher-bash (approved):

PR #483, opened 2024-03-07
fix (app): Handle device auth state change in-between onboarding

ujjwal-cyph:
Fixes https://app.clickup.com/t/9002019994/PRF-7078


rohanagarwal94:
Task linked: [PRF-7078 On cysync in onboarding flow if device connected in bootloader mode with firmware erased then after update device auth is getting skipped](https://app.clickup.com/t/9002019994/PRF-7078)

cypher-bash (approved):

PR #485, opened 2024-03-12
fix (evm): Domain or/and Message only typedata msg signing added

ParnikaGupta9:


apps/evm_family/evm_typed_data_helper.c
```diff
@@ -178,7 +178,14 @@ bool evm_get_typed_struct_data_digest(
   uint8_t *data = NULL;
   uint16_t data_size = 0, offset = 0;
 
-  data_size = sizeof(ETH_SIGN_TYPED_DATA_IDENTIFIER) - 1 + HASH_SIZE * 2;
+  data_size = sizeof(ETH_SIGN_TYPED_DATA_IDENTIFIER) - 1;
+  if (0 < (unsigned int)(typed_data->domain).size) {
```
ujjwal-cyph, line 182: > (unsigned int)

type conversion shouldn't be necessary

apps/evm_family/evm_typed_data_helper.c
```diff
@@ -187,9 +194,17 @@ bool evm_get_typed_struct_data_digest(
          sizeof(ETH_SIGN_TYPED_DATA_IDENTIFIER) - 1);
 
   offset += sizeof(ETH_SIGN_TYPED_DATA_IDENTIFIER) - 1;
-  eip712_status = hash_struct(&(typed_data->domain), data + offset);
+  if (0 < typed_data->domain.size) {
+    eip712_status = hash_struct(&(typed_data->domain), data + offset);
+  } else {
+    eip712_status = 0;
+  }
   offset += HASH_SIZE;
```
ujjwal-cyph, line 202: please confirm if this increment should happen conditionally or not. this might result in an invalid final digest.

apps/evm_family/evm_typed_data_helper.c
```diff
@@ -187,9 +193,17 @@ bool evm_get_typed_struct_data_digest(
...
-  eip712_status = hash_struct(&(typed_data->domain), data + offset);
-  offset += HASH_SIZE;
-  eip712_status |= hash_struct(&(typed_data->message), data + offset);
+  if (0 < typed_data->domain.children_count) {
+    eip712_status = hash_struct(&(typed_data->domain), data + offset);
+    offset += HASH_SIZE;
+    if (0 < typed_data->message.children_count) {
+      eip712_status |= hash_struct(&(typed_data->message), data + offset);
+    }
+  } else if (!strncmp(UI_TEXT_EIP712_DOMAIN_TYPE,
```
ujjwal-cyph, line 202:
```suggestion
  } else if (0 == strncmp(UI_TEXT_EIP712_DOMAIN_TYPE,
```

this is clearer

ujjwal-cyph: also, what is this `else...if` case? when & how could it pop up in a real-world scenario?

ParnikaGupta9: for test case 1: [empty data ](https://github.com/trezor/trezor-firmware/blob/main/common/tests/fixtures/ethereum/sign_typed_data.json)

[https://github.com/trezor/trezor-firmware/blob/54468abd89af62f4a7395398eb358d5fe687430c/core/src/apps/ethereum/sign_typed_data.py#L93](# Setting the primary_type to "EIP712Domain" is technically in spec
# In this case, we ignore the "message" part and only use the "domain" part
# https://ethereum-magicians.org/t/eip-712-standards-clarification-primarytype-as-domaintype/3286)

ujjwal-cyph: please add this as a comment.

VeronicaGupta: added

VeronicaGupta (commented):
added

ujjwal-cyph (approved):
LGTM

PR #536, opened 2024-08-20
feat: Add session apis in core

TejasvOnly:


common/interfaces/card_interface/nfc.c
vrockz747, line 1: do we need to document this core nfc changes? @TejasvOnly

common/interfaces/desktop_app_interface/usb_event.c
```diff
@@ -161,12 +169,13 @@ static void clear_msg_context() {
...
 static core_error_type_t get_core_req_type(usb_core_msg_t msg,
                                            size_t *request_type) {
-  core_msg_t core_msg_p = CORE_MSG_INIT_ZERO;
+  memzero(&core_msg_p, sizeof(core_msg_t));
   pb_istream_t stream = pb_istream_from_buffer(msg.buffer, msg.size);
   core_error_type_t status = CORE_INVALID_MSG;
   // invalid buffer ref, 0 size & decode failure are error situation
   if (false == pb_decode(&stream, CORE_MSG_FIELDS, &core_msg_p) ||
       NULL == msg.buffer || 0 == msg.size) {
+    // TODO: ADD PBERROR LOG
```
vrockz747, line 178:
```suggestion
        LOG_ERROR("Decoding failed: %s\n", PB_GET_ERROR(&stream));
```

common/interfaces/desktop_app_interface/usb_event.c
```diff
@@ -277,8 +291,17 @@ bool usb_get_event(usb_event_t *evt) {
         populate_version_list(&resp);
         send_app_version_list_to_host(&resp);
         reset_event_obj(&usb_event);
+
+      } else if (CORE_MSG_SESSION_START_TAG == request_type) {
+        core_session_parse_message(&core_msg_p);
```
vrockz747, line 296:
```suggestion
        core_session_start_parse_message(&core_msg_p);
```
the function name is too generic, though its parsing only session start queries

src/card_operations/card_fetch_data.h
```diff
@@ -0,0 +1,55 @@
+/**
+ * @file    card_fetch_data.h
+ * @author  Cypherock X1 Team
+ * @brief   API for deleting wallet share from a card
```
vrockz747, line 4: update file brief

src/card_operations/card_fetch_data.h
```diff
@@ -0,0 +1,55 @@
...
+/*****************************************************************************
+ * TYPEDEFS
+ *****************************************************************************/
+#pragma pack(push, 1)
+typedef struct {
+  uint8_t plain_data[PLAIN_DATA_SIZE];
+  uint16_t plain_data_size;
+  uint8_t encrypted_data[ENCRYPTED_DATA_SIZE];
+  uint16_t encrypted_data_size;
+  // is private
```
vrockz747, line 43: pending implementation?

vrockz747 (changes requested):

common/core/core_api.h
```diff
@@ -60,4 +60,10 @@ void send_core_error_msg_to_host(uint32_t which_error);
...
 void send_app_version_list_to_host(
     const core_app_version_result_response_t *version_resp);
-#endif /* CORE_API_H */
\ No newline at end of file
+
+void send_core_session_start_response_to_host(const uint8_t *payload);
+
+void send_core_session_start_ack_to_host();
+
+void send_core_session_close_response_to_host();
```
vrockz747, line 68: we need to document functions before final merge

common/core/core_session.h
```diff
@@ -0,0 +1,83 @@
...
+  uint8_t packet[SESSION_PACKET_SIZE];
+
+} session_config_t;
+#pragma pack(pop)
+
+extern session_config_t session;
+
+void core_session_parse_message(const core_msg_t *core_msg);
+
+void core_session_clear_metadata();
```
vrockz747, line 81: need to add function docs before final merge

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
+/**
+ * @file    session_utils.c
```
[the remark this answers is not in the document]
TejasvOnly: It seems `@file` parameter is redundant as dioxygen can infer the file name without this. Let's just remove this from future files, I'll remove it here as well.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+ * @author  Cypherock X1 Team
+ * @brief   Definition of the session utility functions
+ *          This file defines the functions used to create and manage the
+ *          session, send authentication requests and verify the responses.
+ *
```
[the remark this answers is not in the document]
TejasvOnly: Updated

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+static bool verify_session_signature(uint8_t *payload,
+                                     size_t payload_size,
+                                     uint8_t *signature) {
+  uint8_t hash[32] = {0};
+  sha256_Raw(payload, payload_size, hash);
+
+  uint8_t status = ecdsa_verify_digest(
+      &nist256p1, session.derived_server_public_key, signature, hash);
+
+  return (status == 0);
```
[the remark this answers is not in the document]
TejasvOnly: Yoda notation was never mandatory, it was an old construct that is rendered obsolete with better linter and compiler tools, It's safe to not follow it, the same is mentioned in the coding guidelines. However, I think this line can still be improved, I'll update it.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+}
+
+static bool derive_session_iv() {
```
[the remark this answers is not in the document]
TejasvOnly: Right, we can wrap them with the functions which actually use their values. I'll update this.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+/*****************************************************************************
+ * STATIC VARIABLES
+ *****************************************************************************/
+
+/*****************************************************************************
+ * GLOBAL VARIABLES
+ *****************************************************************************/
+const ecdsa_curve *curve;
+const uint32_t session_key_rotation[2] = {6, 7};
+session_config_t CONFIDENTIAL session = {0};
```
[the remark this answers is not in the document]
TejasvOnly: I don't see how `session_data` is better than `session`, we can discuss this if you want

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+  sha256_Raw(session_iv, 2 * SESSION_PUB_KEY_SIZE, session.session_iv);
+
+  return true;
+}
+
+static bool derive_session_key() {
+  if (ecdh_multiply(curve,
+                    session.device_random,
+                    session.server_random_public,
+                    session.session_key) != 0) {
```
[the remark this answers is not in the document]
TejasvOnly: I don't see how yoda notation would even help in this case. Function calls are not assignable is C.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+    LOG_ERROR("ERROR: Session key not generated");
+    return false;
+  }
+  return true;
+}
+
+static void session_append_signature(uint8_t *payload, size_t payload_size) {
+  uint8_t hash[32] = {0};    // hash size reference taken from the atecc_sign
```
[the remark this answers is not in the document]
TejasvOnly: The atecc utils aren't very well documented, so I've left the comment so we can update this in future.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+  return true;
+}
+
+// TODO: convert this into macro
+static void uint32_to_uint8_array(uint32_t value, uint8_t arr[4]) {
+  arr[0] = (value >> 24) & 0xFF;    // Extract the highest byte
+  arr[1] = (value >> 16) & 0xFF;    // Extract the second highest byte
+  arr[2] = (value >> 8) & 0xFF;     // Extract the second lowest byte
+  arr[3] = value & 0xFF;            // Extract the lowest byte
+}
```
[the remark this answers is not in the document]
TejasvOnly: Won't be included in this pr, this is a project wide change, we have this sort of function at multiple places.

common/core/core_session.c
```diff
@@ -0,0 +1,377 @@
...
+
+static void initiate_request(void) {
+  uint8_t payload[SESSION_BUFFER_SIZE] = {0};
+  if (!session_send_device_key(payload)) {
+    // TODO: Error Handling
+    LOG_ERROR("xxec %d", __LINE__);
+    comm_reject_invalid_cmd();
+    clear_message_received_data();
```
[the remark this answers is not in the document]
TejasvOnly: We already have basic error handling, This todo is specifically for more specific error handling in future.

common/interfaces/desktop_app_interface/usb_event.c
TejasvOnly: This should be covered as separate project wide change

common/interfaces/card_interface/nfc.c
TejasvOnly: There's a [pr](https://github.com/Cypherock/x1_wallet_firmware/pull/525) related to this, I'll leave a TODO for now, will need to update it later.

common/core/core_session.c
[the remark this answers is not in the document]
vrockz747: I see, although apart from safety concerns, i feel we should make this mandatory as it more easier to read and has already been followed in our repo?

common/core/core_session.c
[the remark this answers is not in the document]
vrockz747: regarding this i actually referred from already implemented examples of such cases and currently it seems we are following yoda notation. We can change that from now on i think.

common/core/core_session.c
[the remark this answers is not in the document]
vrockz747: makes sense, but we could remove redundant comments, the function seems gpt generated and too verbose.
```suggestion
// TODO: convert this into macro
static void uint32_to_uint8_array(uint32_t value, uint8_t arr[4]) {
  arr[0] = (value >> 24) & 0xFF;
  arr[1] = (value >> 16) & 0xFF; 
  arr[2] = (value >> 8) & 0xFF; 
  arr[3] = value & 0xFF;
}
```

common/interfaces/desktop_app_interface/usb_event.c
vrockz747: yes, we could create a ticket regarding this and add in comments for reference.

common/core/core_session.c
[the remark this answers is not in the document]
vrockz747: alright, `session` seems fine too, on second thought.

common/core/core_session.c
```diff
@@ -0,0 +1,482 @@
...
+/**
+ * @brief Receives and processes the server's key as part of the session
+ * initialization.
+ *
+ * @param server_message The buffer containing the server's message, which
+ * includes the server's random public key, session age, and device ID.
+ * @return true if the server's key is successfully received and processed.
+ * @return false otherwise.
+ */
+static bool session_receive_server_key(uint8_t *server_message);
```
vrockz747, line 107:
```suggestion
static bool session_receive_server_key(const uint8_t *server_message);
```
the input buffer remains unchanged

vrockz747, line 107: We also need to preserve the order protoypes with that of their definition.
for example `session_receive_server_key` depends on `derive_server_public_key` so its declaration is expected after `derive_server_public_key`. 
Same for other functions

common/core/core_session.c
```diff
@@ -0,0 +1,482 @@
...
+                                    curve_point random_public_point);
+
+/**
+ * @brief Sends the device key as part of the session initialization.
+ *
+ * @param payload The buffer to store the payload data.
+ * @return true if the device key is successfully sent.
+ * @return false otherwise.
+ */
+static bool session_send_device_key(uint8_t *payload);
```
vrockz747, line 188:
```suggestion
static bool session_generate_device_key(uint8_t *payload);
```
the function is not sending the keys but only storing it in the ip buffer,
also update the description accordingly

vrockz747 (changes requested):

common/core/core_session.c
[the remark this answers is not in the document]
TejasvOnly: removed the function entirely.

common/core/core_session.c
TejasvOnly: I don't think this is something that's required or even maintainable, I've moved the order around for the case you mentioned, but I don't think this is something we should even consider.

vrockz747: okay, i wanted to discuss that too actually, we can do that later

vrockz747 (approved):

PR #541, opened 2024-08-24
feat(app): Add seed based wallet auth

vrockz747:


apps/inheritance_app/inheritance_auth_wallet.c
```diff
@@ -118,32 +118,49 @@ static bool verify_auth_wallet_inputs() {
...
+                             ERROR_DATA_FLOW_INVALID_DATA);
+      delay_scr_init(ui_text_inheritance_wallet_auth_fail, DELAY_TIME);
+      return false;
+    }
+    memcpy((void *)auth->entropy,
+           msgs[0].encrypted_data,
+           msgs[0].encrypted_data_size);
+    auth->entropy_size = msgs[0].encrypted_data_size;
+  } else {
+    uint8_t seed[64] = {0};
```
TejasvOnly, line 140: Use a macro for `64`

TejasvOnly (changes requested):

apps/inheritance_app/inheritance_auth_wallet.c
vrockz747: we dont have a global macro for seed, should i keep the scope of macro within inheritance?

TejasvOnly (approved):

PR #551, opened 2024-09-06
fix: Refactor encryption and decryption functions

TejasvOnly:
- sanatize functions to properly handle errors
- created a new static function to separate concerns


src/card_operations/card_fetch_encrypt_data.c
```diff
@@ -132,45 +187,12 @@ card_error_type_e card_fetch_encrypt_data(const uint8_t *wallet_id,
     }
 #endif
 
-    if (CARD_OPERATION_SUCCESS == card_data.error_type) {
+    if (card_data.error_type == CARD_OPERATION_SUCCESS) {
```
vrockz747, line 190: i think this check is irrelevant, cause `card_fetch_wallet_name()` is taking `card_data` as value and its failure state is handled by its return value itself.

vrockz747 (changes requested):

vrockz747 (approved):

PR #553, opened 2024-09-08
chore(app): Add encryption error handling

vrockz747:


apps/inheritance_app/inheritance_encrypt_data.c
```diff
@@ -284,12 +333,58 @@ STATIC inheritance_encryption_context_t *encryption_context = NULL;
 /*****************************************************************************
  * STATIC FUNCTIONS
  *****************************************************************************/
+static void encryption_set_defaults() {
+  encryption_error.type = ENCRYPTION_ERROR_DEFAULT;
+  encryption_error.flow = ENCRYPTION_INIT_FLOW;
```
TejasvOnly, line 338: Use Macros

apps/inheritance_app/inheritance_encrypt_data.c
```diff
@@ -334,6 +429,7 @@ STATIC bool inheritance_encryption_handle_inititate_query(
       !get_wallet_name_by_id(query->encrypt.initiate.wallet_id,
                              (uint8_t *)wallet_name,
                              inheritance_send_error)) {
+    SET_ERROR_FLOW_TAG(ENCRYPTION_INIT_FLOW);
```
TejasvOnly, line 432: Shouldn't we put these at the start of every function? rather than only on error?

TejasvOnly (changes requested):

apps/inheritance_app/inheritance_encrypt_data.c
vrockz747: we could do that actually, and that would make error handling more neat. will update this

TejasvOnly (approved):

PR #556, opened 2024-09-18
fix: Fix inheritance session issues

TejasvOnly:


common/libraries/crypto/bip32.c
```diff
@@ -757,75 +809,65 @@ int hdnode_nem_decrypt(const HDNode *node, const ed25519_public_key public_key,
...
+        node->curve->params, node->private_key, digest, sig, pby, is_canonical);
   } else if (node->curve == &curve25519_info) {
-    return 1;  // signatures are not supported
+    return 1;    // signatures are not supported
   } else {
     return hdnode_sign(node, digest, 32, 0, sig, pby, is_canonical);
   }
 }
 
-int hdnode_get_shared_key(const HDNode *node, const uint8_t *peer_public_key,
```
vrockz747, line 802: why is this function removed?

common/libraries/crypto/bip32.c
```diff
@@ -757,75 +809,65 @@ int hdnode_nem_decrypt(const HDNode *node, const ed25519_public_key public_key,
...
-    }
-    curve25519_scalarmult(session_key + 1, node->private_key,
-                          peer_public_key + 1);
-    *result_size = 33;
-    return 0;
-  } else {
-    *result_size = 0;
-    return 1;  // ECDH is not supported
-  }
-}
```
vrockz747, line 825: is this delete intentional? why is this deleted?

TejasvOnly: Not used anywhere, seems like an old function which was never used.

vrockz747 (approved):

PR #562, opened 2024-09-23
fix: Chunking for empty messages

TejasvOnly:


apps/inheritance_app/inheritance_encrypt_data.c
```diff
@@ -364,7 +364,7 @@ static bool decode_inheritance_plain_data(
     const uint8_t *data,
     uint16_t data_size,
     inheritance_encrypt_data_with_pin_plain_data_structure_t *plain_data) {
-  if (NULL == data || NULL == plain_data || 0 == data_size) {
```
vrockz747, line 367: `data_size` parameter is the size of the pb encoded bytes of encrypted_data_structure, it cannot be zero, if an empty message is incoming the value of `encrypted_data.data.size` would be 0

apps/inheritance_app/inheritance_decrypt_data.c
```diff
@@ -298,7 +298,7 @@ static bool decode_inheritance_encrypted_data(
     uint16_t data_size,
     inheritance_decrypt_data_with_pin_encrypted_data_structure_t
         *encrypted_data) {
-  if (NULL == data || NULL == encrypted_data || 0 == data_size) {
```
vrockz747, line 301: same here as described in `ecryption file`, but also in case of silver plan do we want to encrypt empty struct? i.e i assumed the value of plain_data_count from cysync would be zero so only the PIN would be encrypted, therefore removing the need of decrypting empty plain data at all.

vrockz747 (changes requested):

PR #565, opened 2024-09-27
fix(ci): Add virtual env creation

vrockz747:


TejasvOnly (changes requested):
I think it's better to move python setup to another step using setup-python action or something similar

TejasvOnly:
Picked up in #572

PR #571, opened 2024-10-01
fix: Inheritance bugs

vrockz747:
Fixes:
PRF-7597: https://app.clickup.com/t/9002019994/PRF-7597
PRF-7612: https://app.clickup.com/t/9002019994/PRF-7612
PRF-7603: https://app.clickup.com/t/9002019994/PRF-7603
PRF-7545: https://app.clickup.com/t/9002019994/PRF-7545

### DIFF
--- a/apps/inheritance_app/inheritance_auth_wallet.c
+++ b/apps/inheritance_app/inheritance_auth_wallet.c
@@ -287,7 +287,7 @@ static bool auth_wallet_get_seed_entropy() {
 
 static bool auth_wallet_pair_card() {
   card_error_type_e status = single_card_pair_operation(
-      (char *)ui_text_tap_the_card, ui_text_place_card_below);
+      (char *)ui_text_tap_a_card, ui_text_place_card_below);
   if (status != CARD_OPERATION_SUCCESS) {
     SET_ERROR_TYPE(AUTH_WALLET_PAIRING_ERROR);
     return false;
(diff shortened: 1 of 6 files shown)


src/card_operations/card_fetch_wallet_list.c
```diff
@@ -247,7 +246,6 @@ bool card_fetch_wallet_name(const uint8_t *wallet_id, char *wallet_name) {
   }
 
   if (0 == strlen(wallet_name)) {
-    delay_scr_init(ui_text_wallet_doesnt_exists_on_this_card, DELAY_TIME);
```
TejasvOnly, line 250: Instead of removing the ui enirely, check if an abort is recieved and return without displaying UI

apps/inheritance_app/inheritance_auth_wallet.c
```diff
@@ -287,7 +287,7 @@ static bool auth_wallet_get_seed_entropy() {
 
 static bool auth_wallet_pair_card() {
   card_error_type_e status = single_card_pair_operation(
-      (char *)ui_text_tap_the_card, ui_text_place_card_below);
```
TejasvOnly, line 290: This was the only usage for this text, we can remove it from `constants_text.c/h` as well

TejasvOnly (changes requested):

src/card_operations/card_fetch_wallet_list.c
vrockz747: I see, the p0 check is already added above, but i think it was failing as earlier `card_data` error hanling had bugs.

TejasvOnly (approved):

PR #588, opened 2024-11-08
chore: Add core session error tag

vrockz747:


TejasvOnly (changes requested):
LGTM!, update the submodule

PR #591, opened 2024-11-14
fix: Set proper flow status in ecnryption flow

TejasvOnly:
Fixes: https://app.clickup.com/t/9002019994/PRF-7750

### DIFF
--- a/apps/inheritance_app/inheritance_encrypt_data.c
+++ b/apps/inheritance_app/inheritance_encrypt_data.c
@@ -663,6 +663,7 @@ static bool encrypt_data(void) {
       status = false;
       break;
     }
+    set_app_flow_status(INHERITANCE_ENCRYPT_DATA_STATUS_PIN_ENTERED);
 
     if (!serialize_message_data()) {
       status = false;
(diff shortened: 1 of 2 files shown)


apps/inheritance_app/inheritance_encrypt_data.c
```diff
@@ -673,6 +674,7 @@ static bool encrypt_data(void) {
       status = false;
       break;
     }
+    set_app_flow_status(INHERITANCE_ENCRYPT_DATA_STATUS_MESSAGE_ENCRYPTED);
```
vrockz747, line 677: This would false update cysync even if msg encrpytion fails.
You can put it inside encrypt_message when it succeeds

apps/inheritance_app/inheritance_encrypt_data.c
```diff
@@ -663,6 +663,7 @@ static bool encrypt_data(void) {
       status = false;
       break;
     }
+    set_app_flow_status(INHERITANCE_ENCRYPT_DATA_STATUS_PIN_ENTERED);
```
vrockz747, line 666: same as encrypt_message

vrockz747 (changes requested):

apps/inheritance_app/inheritance_encrypt_data.c
TejasvOnly: Flow status is idependent of erros

apps/inheritance_app/inheritance_encrypt_data.c
TejasvOnly: Flow status is idependent of erros

vrockz747 (approved):

PR #597, opened 2024-11-21
chore: Update card send errors to host

vrockz747:


apps/inheritance_app/inheritance_main.c
```diff
@@ -124,6 +124,7 @@ void inheritance_main(usb_event_t usb_evt, const void *app_config) {
    * servicing a USB initiated command */
   core_status_set_idle_state(CORE_DEVICE_IDLE_STATE_USB);
 
+  LOG_SWV("%s (%d) - Query:%d\n", __func__, __LINE__, query.which_request);
```
TejasvOnly, line 127: is this intentional?

TejasvOnly (changes requested):
Update doxygen comments for the functions with updated signatures

apps/inheritance_app/inheritance_main.c
vrockz747: Yes, i did not remove it because it was followed in all apps as standard.
is it not recommended?

TejasvOnly (approved):

PR #614, opened 2024-12-11
feat(app): Add starknet gen pubkey

vrockz747:


apps/starknet_app/starknet_helpers.c
```diff
@@ -0,0 +1,220 @@
...
+bool grind_key(const uint8_t *grind_seed, uint8_t *out) {
+  uint8_t key[32] = {0};
+  mpz_t strk_limit;
+  mpz_t strk_key;
+  mpz_t stark_order;
+
+  // Initialize stark_order
+  mpz_init_set_str(
+      stark_order,
+      "0800000000000010FFFFFFFFFFFFFFFFB781126DCAE7B2321E66A241ADC64D2F",
```
TejasvOnly, line 116: Repeated use of start order

apps/starknet_app/starknet_pub_key.c
```diff
@@ -0,0 +1,540 @@
...
+  memcpy(call_data[2], one_bn, STARKNET_BIGNUM_SIZE);
+
+  // get call data hash
+  uint8_t call_data_hash[PEDERSEN_HASH_SIZE] = {0};
+  compute_hash_on_elements(call_data, CALL_DATA_PARAMETER_SIZE, call_data_hash);
+
+  uint8_t starknet_contract_address_bn[STARKNET_BIGNUM_SIZE] = {0};
+  hex_string_to_byte_array(
+      "00000000000000535441524b4e45545f434f4e54524143545"
+      "f41444452455353",    ///< bn
```
TejasvOnly, line 402: use this from a variable or a macro

TejasvOnly (changes requested):
Try to remove any magic numbers you can by using macros, and add proper error handling.

common/libraries/crypto/mpz_operations/mpz_ecdsa.c
```diff
@@ -0,0 +1,288 @@
...
+                         mpz_curve_point *cp2) {
+  mpz_t lambda, inv, xr, yr;
+
+  mpz_init(lambda);
+  mpz_init(inv);
+  mpz_init(xr);
+  mpz_init(yr);
+
+  if (mpz_curve_point_is_infinity(cp1)) {
+    return;
```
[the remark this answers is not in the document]
TejasvOnly: Also make sure that every time you have an early return, you clear the mpzs

common/libraries/crypto/mpz_operations/mpz_ecdsa.c
```diff
@@ -0,0 +1,288 @@
...
+  }
+  if (mpz_curve_point_is_infinity(cp2)) {
+    mpz_curve_point_copy(cp1, cp2);
```
TejasvOnly, line 121: This function re-initializes `cp2`, that will cause dangling pointers if `cp2` is already initialized

vrockz747: noted, will fix it; 
also re-initialization would be causing memory leaks rather than dangling pointers?

common/libraries/crypto/mpz_operations/mpz_ecdsa.c
```diff
@@ -0,0 +1,288 @@
...
+}
+
+void mpz_curve_point_clear(mpz_curve_point *p) {
+  mpz_clear(p->x);
+  mpz_clear(p->y);
+}
+
+// Set cp2 = cp1
+void mpz_curve_point_copy(const mpz_curve_point *cp1, mpz_curve_point *cp2) {
+  mpz_curve_point_init(cp2);
```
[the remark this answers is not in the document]
vrockz747: re-initializations would cause memory leaks, will remove all initializations of mpz vars whose scope is outside the function.

common/libraries/crypto/mpz_operations/mpz_ecdsa.c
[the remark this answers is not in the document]
vrockz747: The convention seems to be followed same as Trezor-crypto's `point-add`.
Also i dont have fundamental knowledge of it, but the operation seem successful for all flows,
like when one point is infinity the non-infinire point is copied to result before return, early return is not indicating a necessary failed operation. Check and tell if it requires further change; my comments are based from trezors implementation.

TejasvOnly (approved):

PR #617, opened 2024-12-14
Feat/starknet/refactored/sign txn

vrockz747:


apps/starknet_app/starknet_sign_txn.c
```diff
@@ -0,0 +1,627 @@
...
+      starknet_txn_context->deploy_txn = &query->sign_txn.txn.deploy_txn;
+    } break;
+
+    default: {
+      // should not reach here;
+      return false;
+    }
+  }
+
+  if (1) {
```
TejasvOnly, line 317: ??

apps/starknet_app/starknet_poseidon.h
```diff
@@ -0,0 +1,128 @@
...
+  (MAX_AMOUNT_BITS + MAX_PRICE_PER_UNIT_BITS)    // Combined offset
+#define L1_GAS_NAME "4c315f474153"    // The constant value for L1_GAS_NAME
+#define L2_GAS_NAME "4c325f474153"
+#define INVOKE_TXN_PREFIX                                                      \
+  { 0x69, 0x6e, 0x76, 0x6f, 0x6b, 0x65 }    // 0x696e766f6b65; 'INKVOKE'
+#define DEPLOY_ACCOUNT_PREFIX                                                  \
+  {                                                                            \
+    0x64, 0x65, 0x70, 0x6c, 0x6f, 0x79, 0x5f, 0x61, 0x63, 0x63, 0x6f, 0x75,    \
+        0x6e, 0x74                                                             \
+  }    // 0x6465706c6f795f6163636f756e74 'DEPLOY_ACCOUNT'
```
TejasvOnly, line 92: Are all of these related to poseidon hash? if not, they shouldn't be in this file

common/libraries/crypto/mpz_operations/mpz_ecdsa.c
```diff
@@ -286,3 +330,106 @@ int bn_bit_length(const mpz_t k) {
...
 }
+
+int starknet_sign_digest(const mpz_curve *curve,
+                         const uint8_t *priv_key,
+                         const uint8_t *digest,
+                         uint8_t *sig) {
+  int i = 0;
+  mpz_curve_point R = {0};
+  mpz_t k, z, randk;
+  mpz_curve_point_init(&R);
```
TejasvOnly, line 341: This is never cleared

TejasvOnly (changes requested):

apps/starknet_app/starknet_poseidon.h
vrockz747: yes, releated to starkent context of poseidon hash, all values are used only in `starknet_poseidon.c`, so moved those private macro section in `starknet_poseidon.c`

apps/starknet_app/starknet_sign_txn.c
vrockz747: 😂 no idea why its there; remenants from previous implementation of app; fixed it

TejasvOnly (approved):
There are still a few magic numbers that could be replaced by macro

TejasvOnly (approved):

PR #618, opened 2025-01-01
feat: Add solana token txn parse

vrockz747:


apps/solana_app/solana_sign_txn.c
```diff
@@ -270,6 +282,16 @@ STATIC bool solana_handle_initiate_query(const solana_query_t *query) {
...
+
+  solana_txn_context->is_token_transfer_transaction =
+      query->sign_txn.initiate.has_token_data;
+
+  // if it is a token transfer transaction, store the token data
+  if (solana_txn_context->is_token_transfer_transaction) {
+    memcpy(&solana_txn_context->token_data,
+           &query->sign_txn.initiate.token_data,
+           sizeof(solana_sign_txn_initiate_token_data_t));
+  }
```
TejasvOnly, line 294: This should be done before user confirmation and show the confirmation message according to this

apps/solana_app/solana_txn_helpers.c
```diff
@@ -198,13 +200,43 @@ int solana_byte_array_to_unsigned_txn(uint8_t *byte_array,
...
+        utxn->instruction.program.transfer.amount =
             U64_READ_LE_ARRAY(utxn->instruction.opaque_data + 4);
         break;
 
       default:
         break;
     }
+  } else {
+    //   Set System instruction address for Token Program
+    hex_string_to_byte_array(SOLANA_TOKEN_PROGRAM_ADDRESS,
```
TejasvOnly, line 212: Use else if and explicitly check if the instruction address is for token program

apps/solana_app/solana_txn_helpers.c
```diff
@@ -236,7 +269,27 @@ int solana_validate_unsigned_txn(const solana_unsigned_txn *utxn) {
         break;
     }
   } else {
-    return SOL_V_UNSUPPORTED_PROGRAM;
+    //   Set System instruction address for Token Program
```
TejasvOnly, line 272: Same here

apps/solana_app/solana_txn_helpers.h
```diff
@@ -68,7 +103,7 @@ enum SOLANA_ERROR_CODES {
 typedef struct solana_transfer_data {
   uint8_t *funding_account;
   uint8_t *recipient_account;
-  uint64_t lamports;
+  uint64_t amount;
```
TejasvOnly, line 106: Why is this changed? The docs linked above have this feild as `lamports`, so it is better to keep the names similar  to the reference as much as we can.

TejasvOnly (changes requested):

apps/solana_app/solana_txn_helpers.h
vrockz747: lamports is specific unit with 9 decimals; amount is generic as unit would be decided based on token address

TejasvOnly (approved):

PR #624, opened 2025-01-23
feat: Adds compute budget program support in solana

muzaffarbhat07:


apps/solana_app/solana_txn_helpers.h
```diff
@@ -146,10 +171,12 @@ typedef struct solana_unsigned_txn {
...
 
-  uint16_t instructions_count;    // deserialization only supports max 2
+  uint16_t instructions_count;    // deserialization only supports max 4
                                   // instructions: create account and transfer
-  solana_instruction instruction[2];    ///< Expects max 2 instructions
-  uint8_t transfer_instruction_index;
+  solana_instruction instruction[4];     ///< Expects max 4 instructions
+  uint8_t transfer_instruction_index;    // Expects only 1 transfer instruction
+  uint32_t compute_unit_limit;           // To calculate priority fee
+  uint64_t compute_unit_price_micro_lamports;
```
TejasvOnly, line 179: These two aren't a part of an unsigned transaction, these should be in their own instruction

apps/solana_app/solana_txn_helpers.h
```diff
@@ -146,10 +171,12 @@ typedef struct solana_unsigned_txn {
 
   uint8_t *blockhash;
 
-  uint16_t instructions_count;    // deserialization only supports max 2
+  uint16_t instructions_count;    // deserialization only supports max 4
                                   // instructions: create account and transfer
-  solana_instruction instruction[2];    ///< Expects max 2 instructions
-  uint8_t transfer_instruction_index;
+  solana_instruction instruction[4];     ///< Expects max 4 instructions
```
TejasvOnly, line 176: Add a todo to make this completely dynamic in future

apps/solana_app/solana_txn_helpers.c
```diff
@@ -148,7 +148,7 @@ int solana_byte_array_to_unsigned_txn(uint8_t *byte_array,
   utxn->blockhash = byte_array + offset;
   offset += SOLANA_BLOCKHASH_LENGTH;
 
-  // Instructions: Currently expecting count to be 1 or 2. TODO: Handle batch
+  // Instructions: Currently expecting count to be 3 or 4. TODO: Handle batch
```
TejasvOnly, line 151: Does this mean old versions of cysync won't be able to use the newer firmware? If yes, we don't want this behaviour, we'd like to mantain backwards compatibilty therefore the instruction count of 1 should always be suppored.

TejasvOnly (changes requested):
You seem to have pushed your old branch as it is, please always create a new branch from the base branch before commiting your code. or at least look at the diff and make sure only the new commits are visible.

TejasvOnly (approved):

PR #631, opened 2025-03-08
feat: Added sign txn functionality to constellation app

muzaffarbhat07:


apps/constellation_app/constellation_helpers.c
TejasvOnly, line 1: Use the `n` variants of string manipulation functions in c

src/constant_texts.h
```diff
@@ -56,6 +56,7 @@
 #define UI_TEXT_PIN "PIN\n %s"
 #define UI_TEXT_VERIFY_DESTINATION_TAG "Verify Destination Tag\n%lu"
 #define UI_TEXT_VERIFY_PRIORITY_FEE "Verify Priority Fee\n%s\n%s"
+#define UI_TEXT_VERIFY_FEE "Verify Fee\n%s\n%s"
```
TejasvOnly, line 59: Is there a problem with the old string that every other coin uses?

apps/constellation_app/constellation_txn.c
```diff
@@ -0,0 +1,445 @@
...
+  sha256_Raw(serialized_txn, serialized_txn_len, sha256_digest);
+
+  char sha256_hex_str[SHA256_DIGEST_LENGTH * 2 + 1] = "";
+  byte_array_to_hex_string(sha256_digest,
+                           sizeof(sha256_digest),
+                           sha256_hex_str,
+                           sizeof(sha256_hex_str));
+
+  uint8_t sha512_digest[SHA512_DIGEST_LENGTH] = {0};
+  sha512_Raw((uint8_t *)sha256_hex_str, strlen(sha256_hex_str), sha512_digest);
```
TejasvOnly, line 378: strlen returns the length of the string contained, you just want to use `SHA256_DIGEST_LENGTH * 2 + 1` here

TejasvOnly (changes requested):

apps/constellation_app/constellation_txn.c
muzaffarbhat07: We want the string length only

src/constant_texts.h
muzaffarbhat07: Btc uses its own string. Other coins don't use except evm which uses the combination of 2 strings which I find weird and inconsistent with other data we are verifying

apps/constellation_app/constellation_helpers.c
muzaffarbhat07: done

apps/constellation_app/constellation_txn.c
TejasvOnly: Still replace it with `strnlen`

TejasvOnly (approved):

PR #632, opened 2025-03-18
feat: Adds sign msg functionality to constellation app

muzaffarbhat07:


apps/constellation_app/constellation_helpers.c
```diff
@@ -234,4 +249,47 @@ void serialize_txn(const constellation_transaction_t *txn,
...
+    buffer = (buffer << 6) | value;
+    bits_collected += 6;
+    if (bits_collected >= 8) {
+      bits_collected -= 8;
+      decoded[j++] = (char)((buffer >> bits_collected) & 0xFF);
+    }
+  }
+  decoded[j] = '\0';
+  return decoded;
 }
```
TejasvOnly, line 295: We already have this function, no need to reimplement it. And even if we had to, don't use malloc for returning values always prefer an output pointer argument.

apps/constellation_app/constellation_sign_msg.c
```diff
@@ -0,0 +1,559 @@
...
+            ui_text_title, decoded_json, constellation_send_error);
+
+        free(decoded_json);
+        decoded_json = NULL;
+      }
+    } break;
+
+    default: {
+      result = core_confirmation(UI_TEXT_BLIND_SIGNING_WARNING,
+                                 constellation_send_error);
```
TejasvOnly, line 402: Do we want to send error in blind signing?

apps/constellation_app/constellation_sign_msg.c
```diff
@@ -0,0 +1,559 @@
...
+
+  char length_string[20] = "";
+  size_t length_string_len =
+      snprintf(length_string, sizeof(length_string), "%zu\n", msg_len);
+
+  size_t total_len = prefix_len + length_string_len + msg_len;
+
+  strcpy(prefixed_msg_data, prefix);
+  strcat(prefixed_msg_data, length_string);
+  strcat(prefixed_msg_data, (const char *)ctx->msg_data);
```
TejasvOnly, line 428: use `strnlen`, `strncpy`, and `strncat` instead. and you should take all the sizes that are important in the parameter instead. This function is vulnerable to various issues/edge cases.

TejasvOnly (changes requested):

apps/constellation_app/constellation_sign_msg.c
TejasvOnly: @muzaffarbhat07 missed this

TejasvOnly (approved):

PR #689, opened 2025-09-17
feat: Add hyperliquid support

0xcuriosity:


apps/evm_family/hyperliquid/hyperliquid_app.c
```diff
@@ -0,0 +1,154 @@
...
+    .lunit_name = "HYPE",
+    .name = "Hyperliquid",
+    .chain_id = 999,
+
+    // whitelisted contracts
+    .is_token_whitelisted = is_token_whitelisted,
+};
+
+static const cy_app_desc_t hyperliquid_app_desc = {
+    .id = 25,
```
muzaffarbhat07, line 123: app id is already taken by stellar app. Use a different id

0xcuriosity: made changes

PR #714, opened 2026-02-23
feat: get address along with btc get pubkey

muzaffarbhat07:


rozpuch1987:
https://www.esma.europa.eu/press-news/esma-news/esma-and-european-commission-publish-guidance-non-mica-compliant-arts-and-emts

